PR #5, opened 2018-02-23
Add chapter about unit test's style

tuommaki:
Currently there are variety of styles in unit tests in various projects.
To help test readability and maintenance, common style should be
achieved for unit tests.

### DIFF
--- a/go/general_style_guide.md
+++ b/go/general_style_guide.md
@@ -404,6 +404,5 @@ func Test_ClusterID(t *testing.T) {
 
 **Cons:**
 
-- description writing might occasionally require creativity
 - description gets normalized when transformed to test name -> search in code
   doesn't match the one in `go test` output.
--- a/go/general_style_guide.md
+++ b/go/general_style_guide.md
@@ -320,8 +320,14 @@ case in table-driven test should be executed as a sub-test with
 is then easier and one can execute a single specific test case from within table
 if needed.
 
-In table-driven tests the fields of test case struct must be private because
-semantically it doesn't make sense to expose them.
+In table-driven tests the fields of test case struct must be exported for
+improved readability.
+
+Description field must start with `case n` where `n` is the index of the test
+starting from 0.
+
+Error matchers `func(error) bool` should be used to assert errors. This makes it
+easier to match errors returned from other packages.
 
 All test validations must be included in `t.Run(...) { }` block as technically
 each test case in table is a single test.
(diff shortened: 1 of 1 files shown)


go/general_style_guide.md
```diff
@@ -304,3 +304,115 @@ import (
 - imports are easier to read
 
 **Cons:**
+
+
+## Unit Tests
+
+Unit tests should be written when it makes sense. 100% coverage is not something
```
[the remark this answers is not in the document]
tuommaki: Reworded a bit. Do you think this is now covered or would you have suggestions for alternative?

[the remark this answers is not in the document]
fgimenez: Better, thanks! I don't totally agree with your view but that's ok, we can use tests for different things :)

go/general_style_guide.md
```diff
@@ -304,3 +304,123 @@ import (
...
+			case err == nil && tt.ErrorMatcher == nil: // correct; carry on
+			case err != nil && tt.ErrorMatcher != nil:
+				if !tt.ErrorMatcher(err) {
+					t.Fatalf("error == %#v, want matching", err)
+				}
+			case err != nil && tt.ErrorMatcher == nil:
+				t.Fatalf("error == %#v, want nil", err)
+			case err == nil && tt.ErrorMatcher != nil:
+				t.Fatalf("error == nil, want non-nil")
+			}
```
kopiczko: I think this can be simplified to

```go
switch {
case err == nil && tt.ErrorMatcher == nil:
	// correct; carry on
case err != nil && tt.ErrorMatcher == nil:
	t.Fatalf("error == %#v, want nil", err)
case err == nil && tt.ErrorMatcher != nil:
	t.Fatalf("error == nil, want non-nil")
case !tt.ErrorMatcher(err):
	t.Fatalf("error == %#v, want matching", err)
}
```

go/general_style_guide.md
```diff
@@ -304,3 +304,123 @@ import (
...
+			}{
+				foo: "wrong type",
+				bar: -1,
+			},
+			ExpectedClusterID: "",
+			ErrorMatcher:      IsWrongTypeError,
+		},
+	}
+
+	for _, tt := range testCases {
```
kopiczko: I was looking at blogs and examples. I see both `tests` and then `tt` and `testCases` and then `tc`. When we are going use `"case 2: ..."` I think `tc` makes more sense finally.

go/general_style_guide.md
```diff
@@ -304,3 +304,115 @@ import (
...
+sense when facing a bug that could have been prevented by having a test for the
+relevant code in question.
+
+Use table-driven tests instead of separate test methods when possible. Each test
+case in table-driven test should be executed as a sub-test with
+[t.Run](https://golang.org/pkg/testing/#T.Run) as following output of `go test`
+is then easier and one can execute a single specific test case from within table
+if needed.
+
+In table-driven tests the fields of test case struct must be exported for
```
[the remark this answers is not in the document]
kopiczko: I reckon all the blogs, and examples we exchanged have test struct fields unexpected. So even though I like exported more (and it makes more sense to me personally) I'd go with unexported to be more upstream.
Not very important though.

go/general_style_guide.md
tuommaki: :+1:

go/general_style_guide.md
tuommaki: Changed to `tc`.

pipo02mix:
I am a go beginner programmer, but in some go's talks about testing, I have seen they encourage to try to run tests parallel, why do not consider this possibility?

tuommaki:
> I am a go beginner programmer, but in some go's talks about testing, I have seen they encourage to try to run tests parallel, why do not consider this possibility?

I also noticed `t.Parallel()` the other day when browsing Godoc, but I'm not familiar with this and Godoc for this wasn't exhaustive. Do you know how it would behave if it's called both, in the beginning of the actual test method and then inside the sub-test?

In general I think this could be a good practice, but as I'm not personally too familiar with this, I would leave this up to test writer's responsibility to test/add this to implementation. Just like `t.Helper()` should be called appropriately. WDYT?

pipo02mix:
I think in order to use `t.Parallel()` you should not have dependencies between tests so they can run in parallel without issues. So as you said leave it to the test writer. I brought it here because I was wondering if we can improve our CI duration times

tuommaki:
I'm not sure if it has that big effect on CI duration because for all the cases that I've seen, compilation time is longer than running unit tests. Integration tests (maybe the same thing as e2e in our context?) are a different thing, but I'd guess depending on situation there might be external resources involved (part of the reason for experienced slowness?) which don't necessarily allow parallelism.

I think it might be a good idea to leave `t.Parallel()` out from this guide for now and make experimentation whether or not it provides improvements and if it would be a good general advice. WDYT?

go/general_style_guide.md
```diff
@@ -304,3 +304,123 @@ import (
...
+	return customObject.Spec.ID, nil
+}
+```
+
+Example test:
+```go
+
+func Test_ClusterID(t *testing.T) {
+	testCases := []struct {
+		name              string
```
MarcelMue: Dumb question, but since I was trying this out in my own test: How do we want to order the fields in testcases?
`Name` at the top and `errorMatcher` at the bottom and alphabetical in between?

tuommaki: My approach initially was just an instinct more or less based on the order of usage. `name` field appears first in all community versions I have seen so far.

My proposal would be somewhat like this:
1. `name`
2. Input parameters
3. Expected output
4. Possibly validators if needed?
5. Error matcher

MarcelMue: Ack, I would add setup - for example when you need to set up a service or something similar in the fake k8s api. So:
1. `name`
2. Setup config
3. Input parameters
4. Validators
5. Error matchers

tuommaki:
Less and less comments here so I'll merge this now and let's improve it in further iterations later when we find improvements.

PR #6, opened 2018-03-01
Update Kubernetes labels to align with AWS resource tags

rossf7:
I'm proposing we rename `giantswarm.io/cluster-id` to `giantswarm.io/cluster` to align with Kubernetes cloud provider tags.

e.g. `kubernetes.io/cluster/eggs2` Note: they include the cluster ID in the tag so shared resources can be tagged with multiple clusters. I don't think we need to support that.

Also proposed the rename from customer to organization. We should also rename in the CRDs.
https://github.com/giantswarm/giantswarm/issues/2383#issuecomment-369573331 

I can add a section on cloud provider tagging if that is useful. WDYT?

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -5,10 +5,10 @@
 - `app` - value should contain the name of the application. Should be applied
   to every Kubernetes resource associated with an application. This should be
   also used as replicas selector. E.g. `app=kvm-operator`.
-- `giantswarm.io/cluster-id` - value should contain guest cluster ID which this
-  object is part of. E.g. `giantswarm.io/cluster-id=eggs2`.
 - `giantswarm.io/customer-id` - value should contain guest cluster's customer
   ID. `giantswarm.io/customer-id=track-hunter`.
+- `giantswarm.io/cluster` - value should contain guest cluster ID which this
+  object is part of. E.g. `giantswarm.io/cluster=eggs2`.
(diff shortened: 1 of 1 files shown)


kubernetes/annotations_and_labels.md
```diff
@@ -5,10 +5,11 @@
...
   also used as replicas selector. E.g. `app=kvm-operator`.
-- `giantswarm.io/cluster-id` - value should contain guest cluster ID which this
-  object is part of. E.g. `giantswarm.io/cluster-id=eggs2`.
-- `giantswarm.io/customer-id` - value should contain guest cluster's customer
-  ID. `giantswarm.io/customer-id=track-hunter`.
+- `giantswarm.io/cluster` - value should contain guest cluster ID which this
+  object is part of. E.g. `giantswarm.io/cluster=eggs2`.
+- `giantswarm.io/organization` - value should contain guest cluster's
+organization ID as displayed in the front-end
+`giantswarm.io/organization=track-hunter`.
```
kopiczko: Is organization a drop in replacement for customer? We have multiple organizations per customer in happa. Or I confuse things?

rossf7: Yes it's a replacement.

AFAIK the customer naming is from the previous microservices platform. It doesn't make sense now since an installation belongs to a single customer.

kopiczko: Thanks for clarification!

PR #9, opened 2018-03-27
Add giantswarm.io/randomkey

tuommaki:
Randomkey secrets where earlier denoted by clusterKey=<type> which is
somewhat ambiguous. Use giantswarm.io/randomkey for this so it refers to
randomkeys implementation here: https://github.com/giantswarm/randomkeys.

While here, order labels alphabetically.


tuommaki:
You need to start somewhere :) Cluster-operator will now start adding this to `encryptionkey`.

PR #10, opened 2018-03-27
Update finalizers section

MarcelMue:
This is me trying to align `fmt` with the implementation of finalizers in operatorkit.
see https://github.com/giantswarm/giantswarm/issues/1595

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -15,9 +15,13 @@ organization ID as displayed in the front-end
 
 ## Finalizers
 
-- Operator specific finalizers for custom objects shoud be named
-  `OPERATOR_REPO.giantswarm.io/custom-object-cleanup`. E.g.
-  `ingress-operator.giantswarm.io/custom-object-cleanup`.
+- Operatorkit sets a finalizer for custom objects that are watched by the
+  framework named `operatorkit.giantswarm.io/NAME`.
+- `NAME` is configured by the operator and should be the operator name. E.g.
+  `kvm-operator`.
+- If one operator implements multiple frameworks, then `NAME` should include
+  the framework name. E.g `aws-operator-cluster` and
+  `aws-operator-drainer`.
 
 ## Naming
 


kubernetes/annotations_and_labels.md
```diff
@@ -15,9 +15,13 @@ organization ID as displayed in the front-end
...
 
-- Operator specific finalizers for custom objects shoud be named
-  `OPERATOR_REPO.giantswarm.io/custom-object-cleanup`. E.g.
-  `ingress-operator.giantswarm.io/custom-object-cleanup`.
+- Operatorkit sets a finalizer for custom objects that are watched by the
+  framework named `operatorkit.giantswarm.io/NAME`.
+- `NAME` is configured by the operator and should be the operator name. E.g.
+  `kvm-operator`.
+- If one operator implements multiple frameworks, then `NAME` should include
+  the framework name. E.g `aws-operator-cluster` and
```
MarcelMue: I am not really convinced by this naming, but couldn't really think of anything better. The finalizers need to have different names, as they might watch the same resource (which would cause issues if they use the same name).

xh3b4sd: Isn't the scheme something like `domain/key=value`? Multiple frameworks can be denoted by the key/value notation.

xh3b4sd: Would then be these for AWS.

- `operatorkit.giantswarm.io/aws-operator=cluster`
- `operatorkit.giantswarm.io/aws-operator=drainer`

MarcelMue: Yeah, you are right - that is much better. Let me check quickly if that works.

MarcelMue: Well there goes that idea: 
```
metadata.finalizers: Invalid value: "operatorkit.giantswarm.io/aws-operator=cluster": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')
```
Finalizers "arbitrary string values" (https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#finalizers) :(

MarcelMue: That won't work either:
```
 metadata.finalizers[0]: Invalid value: "cluster.aws-operator.operatorkit.giantswarm.io": name is neither a standard finalizer name nor is it fully qualified
```

This works though `cluster.aws-operator.giantswarm.io/operatorkit`. Uncertain if I prefer this though.

xh3b4sd: What a stupid bullshit man. For real. I personally would then prefer this. 

```
- `aws-operator.operatorkit.giantswarm.io/cluster`
- `aws-operator.operatorkit.giantswarm.io/drainer`
```

kopiczko: Then we'll have to pass 2 values to operatorkit. Like ProjectName and ControllerName.

For me both approaches are fine. Both have drawbacks.

xh3b4sd: I see. We have `operatorkit.giantswarm.io` as base and need to append something after the slash. What a disappointing restriction the key/value pair does not work. Then we go with what Marcel suggested in the first place. 

```
- `operatorkit.giantswarm.io/aws-operator-cluster`
- `operatorkit.giantswarm.io/aws-operator-drainer`
```

PR #12, opened 2018-04-09
Add Helm chart formatting guidelines

indyfree:
With respect to a discussion in a recent PR, I propose some helm chart formatting guidelines regarding templating and indentation. This is mostly aligned with upstream helm docs and charts.

This is open for discussion, ideas/proposals welcome.

### DIFF
--- a/helm/formatting_guidelines.md
+++ b/helm/formatting_guidelines.md
@@ -17,8 +17,13 @@ data:
 In comparison to this:
 
 ```yaml
-# No spaces makes it harder to see the templating call
-name: {{.Release.Name}}-configmap
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  # No spaces makes it harder to see the templating call
+  name: {{.Release.Name}}-configmap
+data:
+  myvalue: "Hello World"
 ```
 
 ## Indentation
--- a/helm/formatting_guidelines.md
+++ b/helm/formatting_guidelines.md
@@ -24,7 +24,9 @@ name: {{.Release.Name}}-configmap
 ## Indendation
 
 The indendation of our helm chart yaml files should be 2 spaces.
-Moreover we do not indend yaml lists, hyphens of each list member act as indentation.
+Moreover we do not indend yaml lists, hyphens of the list members act as indentation.
+
+Example:
 ```yaml
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
--- a/helm/formatting_guidelines.md
+++ b/helm/formatting_guidelines.md
@@ -1,4 +1,4 @@
-# Helm Formatting Guidelines
+# Helm Chart Formatting Guidelines
 
 ## Templating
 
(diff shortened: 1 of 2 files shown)


xh3b4sd (approved):
When the team is happy I am happy.

helm/formatting_guidelines.md
```diff
@@ -0,0 +1,54 @@
...
+
+```yaml
+# No spaces makes it harder to see the templating call
+name: {{.Release.Name}}-configmap
+```
+
+## Indendation
+
+The indendation of our helm chart yaml files should be 2 spaces.
+Moreover we do not indend yaml lists, hyphens of the list members act as indentation.
```
MarcelMue: Nitpick `indent` and `indentation` in the line above.

fgimenez: typo, `indent`

fgimenez (approved):
Nice! just one nit, LGTM

helm/formatting_guidelines.md
```diff
@@ -0,0 +1,54 @@
...
+## Templating
+
+When using helm templating calls we include double spaces inside the double brackets
+like this:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-configmap
```
MarcelMue: I think this single line would be enough, you only do a single line in the comparison aswell.

MarcelMue (approved):
Fine with me, we can add more stuff when we get to it.

helm/formatting_guidelines.md
indyfree: I think It's better so see within some context. But added the same example in the comparison for completeness :)

indyfree:
Merging now. We can add to this or change it later if needed. :)

helm/formatting_guidelines.md
```diff
@@ -0,0 +1,59 @@
...
+  # No spaces makes it harder to see the templating call
+  name: {{.Release.Name}}-configmap
+data:
+  myvalue: "Hello World"
+```
+
+## Indentation
+
+The indentation of our helm chart yaml files should be 2 spaces.
+Moreover we do not indent yaml lists, hyphens of the list members act as indentation.
```
kopiczko: According to helm docs we should indent lists: https://docs.helm.sh/developing_charts/#managing-dependencies-with-requirements-yaml

(That's also my personal preference - that's why I mention :P)

kopiczko: Also we should mention what quotation style we use. I guess double quotes, e.g. https://docs.helm.sh/developing_charts/#values-files

indyfree: Yes saw that helm style is different. I personally don't care which style we use, as long as its consistent, which is currently not the case ;)
Just saw that not-intending was mostly prevelent in our main operators repos (e.g. aws, kvm, azure deployments, unfortunatly not rbac, psp) and it seems to be the case most charts upstream https://github.com/kubernetes/charts/tree/master/stable and in kubernetes docs, e.g. https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

indyfree: Yes good call, unfortunatly I already merged it. But your welcome to open another PR :)

PR #13, opened 2018-04-17
add labels for guest cluster services

puja108:
towards https://github.com/giantswarm/giantswarm/issues/2972

(also contains minor typo and line fix to host cluster label section)

Added `app` and `giantswarm.io/type`. Latter is open enough to accommodate introduction of additional types like e.g. a potential `incubation` for managed but not SLA supported services.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -1,20 +1,32 @@
 # Kubernetes Annotations and Labels
 
-## Common Labels
+## Common Labels (for host cluster services)
 
 - `app` - value should contain the name of the application. Should be applied
-  to every Kubernetes resource associated with an application. This should be
-  also used as replicas selector. E.g. `app=kvm-operator`.
+  to every Kubernetes resource associated with an application. This should
+  also be used as replicas selector. E.g. `app=kvm-operator`.
 - `giantswarm.io/cluster` - value should contain guest cluster ID which this
   object is part of. E.g. `giantswarm.io/cluster=eggs2`.
-  operator managing the object. E.g. `giantswarm.io/managed-by=kvm-operator`.
 - `giantswarm.io/managed-by` - value should contain repository name of the
+  operator managing the object. E.g. `giantswarm.io/managed-by=kvm-operator`.
 - `giantswarm.io/organization` - value should contain guest cluster's
-organization ID as displayed in the front-end
-`giantswarm.io/organization=track-hunter`.
+  organization ID as displayed in the front-end
+  `giantswarm.io/organization=track-hunter`.
 - `giantswarm.io/randomkey` - value should contain type of the randomkey stored
   in secret. E.g. `giantswarm.io/randomkey=encryption` for guest API -> etcd
   encryption at rest.
+
+## Common Labels (for guest cluster services)
+
+- `app` - value should contain the name of the application. Should be applied
+  to every Kubernetes resource associated with an application. This should 
+  also be used as replicas selector. E.g. `app=nginx-ingress-controller`.
+  Exceptions can be made to accomodate for adherence to existing selectors
+  upstream.
+- `giantswarm.io/type` - value should be either `system` for core system
+  services (i.e. K8s components) or `managed` for Giant Swarm managed services
+  (i.e. networking, DNS, monitoring, ingress controller, etc.). Latter would be
+  usually managed by `chart-operator`.
 
 ## Finalizers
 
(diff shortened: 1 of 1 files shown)


kubernetes/annotations_and_labels.md
```diff
@@ -1,21 +1,33 @@
...
   in secret. E.g. `giantswarm.io/randomkey=encryption` for guest API -> etcd
   encryption at rest.
 
+## Common Labels (for guest cluster services)
+
+- `app` - value should contain the name of the application. Should be applied
+  to every Kubernetes resource associated with an application. This should 
+  also be used as replicas selector. E.g. `app=nginx-ingress-controller`.
+  Exceptions can be made to accomodate for adherence to existing selectors
+  upstream.
```
kopiczko: Why not just extending the description in previous section?

kubernetes/annotations_and_labels.md
```diff
@@ -1,21 +1,33 @@
...
+- `giantswarm.io/type` - value should be either `system` for core system
+  services (i.e. K8s components) or `managed` for Giant Swarm managed services
+  (i.e. networking, DNS, monitoring, ingress controller, etc.). Latter would be
+  usually managed by `chart-operator`.
```
kopiczko: I don't get when to use `system` and `managed`. My understanding is we should add label to all manifests in k8scloudconfig and all manifests in charts we deploy by chart-operator. But how to distinguish which one is `system` or `managed`?

kubernetes/annotations_and_labels.md
puja108: because I do not want to mix host cluster and guest cluster things, they are different, one the customer sees and the other not

kubernetes/annotations_and_labels.md
puja108: Yeah, maybe we just do a single `managed`? other things might be too confusing

kopiczko: In that case the value carries zero information. So maybe we can establish two thigns:

- Add `ManagedBy` parameter in k8scloudconfig and set `giantswarm.io/managed-by={{ .ManagedBy }}` label to all manifests in the generated cloud config.
- Add `managedBy` value in `values.yaml` in helm charts and set `giantswarm.io/managed-by={{ .Values.managedBy }}` label to all manifests in the chart.

That way we don't have to distinguish between guest/host labels. We still can select manifests by the label name regardless of the value. As a bonus value carries some potentially valuable information.

puja108: What would be the content of `ManagedBy`?

I'm hesitant to mix host and guest cluster as the goal of the labels is different.

For example: In the future we are planning to add services that might not be managed, yet, for services that are still in `incubation` or whatever you want to call it. I want to use this label to communicate to the customer what is giant swarm and managed, and what is giant swarm and maybe not yet managed.

puja108: Currently, I'm not planning on them touching it, but more for communicating the facts of management levels to them on a technical level, and I kind of want to be prepared for different levels of support/management we might have.

kubernetes/annotations_and_labels.md
MarcelMue: I don't see the argument here, why not add all labels on hostcluster aswell? 
It would increase visibility of which things we actively manage and which are `system` on hostcluster IMO.

puja108: I do not argument against adding labels to host cluster, I just argue that host cluster services have a different need and use for labels than guest cluster ones. For host cluster we do have the set above already, but it seems it's not used that much.

We can however talk about if it's enough to have names of operators there or if we need additional labels.

It's just a bit off topic for this PR as I mainly care about solving a customer facing (and support) issue.

kubernetes/annotations_and_labels.md
rossf7: My understanding is we want to easily identify services in guest clusters that will be managed services. As opposed to system services that we run in all guest clusters.

So node-exporter is an example of a system component we always run for our monitoring.
Prometheus could be an example of a future managed service we provide but customers would opt in.

IMO it is useful to label the system components now. The managed services we can add later as we roll them out. The only thing I'm not keen on is the label name `/type` is very generic. I'd prefer `giantswarm.io/component-type` or `/service-type`.

kopiczko: +1 on `/type` being very generic.

puja108: I like `/service-type` and we can see with time if and how we want to split services between things like `managed` and others like `system` maybe

PR #15, opened 2018-05-22
separates go rules into different files

xh3b4sd:
This is something I wanted to do for quite a while now.


MarcelMue (approved):
I like it, thanks <3

PR #16, opened 2018-06-29
adds guest cluster node labels

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918. Also see https://github.com/giantswarm/kubernetesd/pull/388#discussion_r198907607.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -27,15 +27,19 @@
   services (i.e. K8s components) or `managed` for Giant Swarm managed services
   (i.e. networking, DNS, monitoring, ingress controller, etc.). Latter would be
   managed by `chart-operator`.
+
+## Common Labels (for guest cluster nodes)
+
 - `release.giantswarm.io/version` - value should be the used Giant Swarm release
   version, e.g. `2.3.0`. Should be applied to all Kubernetes nodes, regardless
-  their role.
-- `operator.giantswarm.io/version` - value should be the used operator version
-  bundle version, e.g. `1.0.0`. Should be applied to all Kubernetes nodes,
-  regardless their role.
-- `operator.giantswarm.io/provider` - value should be the installation's
-  provider, e.g. `kvm`, `aws`, or `azure`. Should be applied to all Kubernetes
-  nodes, regardless their role.
+  of their role.
+- `NAME.operator.giantswarm.io/version` - value should be the used operator
+  version bundle version, e.g. `1.0.0`. `NAME` might be `kvm` for the
+  `kvm-operator`. Should be applied to all Kubernetes nodes, regardless of their
+  role.
+- `giantswarm.io/provider` - value should be the installation's provider, e.g.
+  `kvm`, `aws`, or `azure`. Should be applied to all Kubernetes nodes,
+  regardless of their role.
 
 ## Finalizers
 
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -33,10 +33,9 @@
 - `release.giantswarm.io/version` - value should be the used Giant Swarm release
   version, e.g. `2.3.0`. Should be applied to all Kubernetes nodes, regardless
   of their role.
-- `NAME.operator.giantswarm.io/version` - value should be the used operator
-  version bundle version, e.g. `1.0.0`. `NAME` might be `kvm` for the
-  `kvm-operator`. Should be applied to all Kubernetes nodes, regardless of their
-  role.
+- `OPERATOR.giantswarm.io/version` - value should be the used operator
+  version bundle version, e.g. `1.0.0`. `OPERATOR` might be `kvm-operator`.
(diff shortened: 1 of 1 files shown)


marians (approved):
No objection

kubernetes/annotations_and_labels.md
```diff
@@ -19,14 +19,23 @@
...
-  to every Kubernetes resource associated with an application. This should 
+  to every Kubernetes resource associated with an application. This should
   also be used as replicas selector. E.g. `app=nginx-ingress-controller`.
   Exceptions can be made to accomodate for adherence to existing selectors
   upstream.
 - `giantswarm.io/service-type` - value should be either `system` for core system
   services (i.e. K8s components) or `managed` for Giant Swarm managed services
   (i.e. networking, DNS, monitoring, ingress controller, etc.). Latter would be
   managed by `chart-operator`.
+- `release.giantswarm.io/version` - value should be the used Giant Swarm release
```
rossf7: I think a separate heading for node labels would be good.

kubernetes/annotations_and_labels.md
```diff
@@ -19,14 +19,23 @@
...
+  version, e.g. `2.3.0`. Should be applied to all Kubernetes nodes, regardless
+  their role.
+- `operator.giantswarm.io/version` - value should be the used operator version
+  bundle version, e.g. `1.0.0`. Should be applied to all Kubernetes nodes,
+  regardless their role.
+- `operator.giantswarm.io/provider` - value should be the installation's
```
rossf7: The version labels LGTM but I think this could just be `giantswarm.io/provider` as to me this isn't operator specific.

kubernetes/annotations_and_labels.md
```diff
@@ -19,14 +19,23 @@
...
   also be used as replicas selector. E.g. `app=nginx-ingress-controller`.
   Exceptions can be made to accomodate for adherence to existing selectors
   upstream.
 - `giantswarm.io/service-type` - value should be either `system` for core system
   services (i.e. K8s components) or `managed` for Giant Swarm managed services
   (i.e. networking, DNS, monitoring, ingress controller, etc.). Latter would be
   managed by `chart-operator`.
+- `release.giantswarm.io/version` - value should be the used Giant Swarm release
+  version, e.g. `2.3.0`. Should be applied to all Kubernetes nodes, regardless
+  their role.
```
MarcelMue: Nitpick: ... regardless `of` their role.

MarcelMue (approved):
LGTM - make Ross happy and I am happy.

kubernetes/annotations_and_labels.md
xh3b4sd: Done.

kubernetes/annotations_and_labels.md
xh3b4sd: Done.

kubernetes/annotations_and_labels.md
xh3b4sd: Done.

rossf7 (approved):
<3

kubernetes/annotations_and_labels.md
```diff
@@ -28,6 +28,19 @@
   (i.e. networking, DNS, monitoring, ingress controller, etc.). Latter would be
   managed by `chart-operator`.
 
+## Common Labels (for guest cluster nodes)
+
+- `release.giantswarm.io/version` - value should be the used Giant Swarm release
+  version, e.g. `2.3.0`. Should be applied to all Kubernetes nodes, regardless
+  of their role.
+- `NAME.operator.giantswarm.io/version` - value should be the used operator
```
xh3b4sd: I added another level for the operator name. My idea is that you can compute the provider operator's version by taking the provider label and read the version label for the provider operator. This makes also more sense for the label structure since there are multiple operators and we should be specific here, just like we made it more straight forward with the provider label.

xh3b4sd (commented):
Please have another look.

kubernetes/annotations_and_labels.md
kopiczko: So in that case I'd say we should stick to general naming scheme: https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md#naming
It would just be `kvm-operator.giantswarm.io/version`

xh3b4sd: I don't like this, but I follow now.

xh3b4sd: Done.

kopiczko: I also think it's ok to have both if that is useful.

```
kvm-operator.giantswarm.io/version
operator.giantswarm.io/version
```

But having something `kvm-operator.operator.giantswarm.io/version` or `kvm.operator.giantswarm.io/version` feels just weird or it's so close to convention we have that I don't see a reason to break it.

kopiczko (approved):

xh3b4sd:
4 approves is good enough for me here. Rolling.

PR #17, opened 2018-07-24
go: add log messages guide

kopiczko:


go/log_messages.md
```diff
@@ -0,0 +1,90 @@
...
+
+```go
+var err error
+
+// Note all messages contain "system fact" which is human readable name for the
+// system fact being discovered. Exception is "cancelling resource" message.
+var systemFact string
+{
+	// Start message. Note "finding out" here.
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding out system fact")
```
kopiczko: I thought about good verb here. Like `get`, `look for`, `figure`. I think `find` works well with our `IsNotFound` matchers.

go/log_messages.md
```diff
@@ -0,0 +1,90 @@
...
+
+	systemFact, err = pkg.FindSystemFact(ctx)
+	if IsNotFound(err) {
+		// Not found log. Note "did not find" here.
+		r.logger.LogCtx(ctx, "level", "debug", "message", "did not find system fact")
+		// Reason log should be separate free form log explaining why
+                // the system fact can't be found.
+		r.logger.LogCtx(ctx, "level", "debug", "message", "waiting for system fact to be created")
```
kopiczko: Added a reason block which serves as an explanation for for the dry log of not being able to find info about the fact.

go/log_messages.md
```diff
@@ -0,0 +1,90 @@
...
+		// Canceling resource log should be separate.
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
```
kopiczko: This is deviation of what we have now in many places but `canceling resource reconciliation for custom object` contains redundant or incorrect information. E.g. when the controller is watching pods we don't reconcile custom objects.

kopiczko (commented):
This is just something I made to achieve the consensus. Not everything here is created as I like it to be, but I tried to find common ground based on what I saw in the reviews and discussion so far. Let's talk!

go/log_messages.md
xh3b4sd: Makes sense. When the whole reconciliation is canceled it should then simply be `canceling reconciliation`.

go/log_messages.md
```diff
@@ -0,0 +1,90 @@
...
+{
+	// Start message. Note "ensured" here. "deletion of" should be present
+	// in case of delete event.
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring [deletion of] managed resource")
+
+	err = pkg.APICall(ctx, systemFact)
+	if IsDeletionInProgress(err) {
+		// Did not ensure log. Note "did not ensure" here. "deletion"
+		// should be present in case of delete event.
+		r.logger.LogCtx(ctx, "level", "debug", "message", "did not ensure [deletion] managed resource")
```
xh3b4sd: `[deletion]` -> `[deletion of]`

xh3b4sd (approved):
This is really cool. I like it a lot.

go/log_messages.md
indyfree: Nitpick: There is an extra space in line 34.

go/log_messages.md
indyfree: For me `canceling resource reconciliation` is the most intuitive. Without the `reconsiliation` it sounds a bit odd to me, but no strong opinions.

indyfree (approved):
Super cool. Makes total sense to align this!

go/log_messages.md
```diff
@@ -0,0 +1,90 @@
+## Log Messages
+
+### Resource Reconciliation
+
+When doing resource reconciliation action we often have to find out the current
+state, and apply changes to bring it towards desired state. This often means
+calling some API and then emitting debug logs about that. We should try to
```
MarcelMue: I think this is the most important part and not limited to reconciliation. We call APIs in other places as well and surrounding our calls with this kind of logging is not only helpful for operators IMO.

MarcelMue (approved):
I like it, we can iterate on this later.

go/log_messages.md
xh3b4sd: @indyfree how do you express the cancelation of the whole loop then? We can either cancel only one resource within the loop or the whole loop.

rossf7 (approved):
LGTM. Will make reading our logs easier which is really cool. <3

go/log_messages.md
kopiczko: Looks like `canceling resource` and `canceling reconciliation` wins here.
We should also write down messages for other context related stuff, like keeping finalizers. But let's make it separate PR.

kopiczko:
Last call. Merging tomorrow morning.
I think we should update with reconciliation canceling keeping finalizers, and other potential context packages logs. This will be in next PR to not make too many discussions at once.

PR #22, opened 2018-08-17
go: add logs before returns

kopiczko:
I feel like this is useful. It allows to detect what was not
accomplished before returning the error. Proper error logging should be
done as it was. This will align logs for canceling reconciliation due to
handled reasons and due to error.

### DIFF
--- a/go/log_messages.md
+++ b/go/log_messages.md
@@ -39,6 +39,9 @@ var systemFact string
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
 		return nil
 	} else if err != nil {
+		// Not found log in error. Note it doesn't log the error on
+                // purpose. this should be done by the caller.
+		r.logger.LogCtx(ctx, "level", "debug", "message", "did not find system fact")
 		return microerror.Mask(err)
 	}
 
@@ -77,6 +80,9 @@ var err error
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
 		return nil
 	} else if err != nil {
+                // Did not ensure log. Note it doesn't log the error on
+                // purpose. this should be done by the caller.
+		r.logger.LogCtx(ctx, "level", "debug", "message", "did not ensure [deletion of] managed resource")
 		return microerror.Mask(err)
 	}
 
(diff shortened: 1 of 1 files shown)


xh3b4sd (commented):
I do not think it is useful to add more logs to error handling. The error is logged already and it becomes kind of self evident that what was intended to be done did not complete or succeed because there was an error. This is not useful.

tuommaki:
I think this is case-by-case. Occasionally it is important to log contextual information when it's not available in the place where error handling happens, but I don't feel this should be a rule. In addition to this, we already have stack trace for errors so this is not needed for execution path tracing.

kopiczko:
I tried to make things consistent. On error we just miss the `did not` log.
My thinking is we should either have it in both places like:

```go
	if IsNotFound(err) {
		r.logger.LogCtx(ctx, "level", "debug", "message", "did not find system fact")
		r.logger.LogCtx(ctx, "level", "debug", "message", "waiting for system fact to be created")
		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
		return nil
	} else if err != nil {
		r.logger.LogCtx(ctx, "level", "debug", "message", "did not find system fact")
		return microerror.Mask(err)
	}
```

Or drop it completely like:

```go
	if IsNotFound(err) {
		r.logger.LogCtx(ctx, "level", "debug", "message", "waiting for system fact to be created")
		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
		return nil
	} else if err != nil {
		return microerror.Mask(err)
	}
```

That would be consistent.

tuommaki:
In that specific example, I like the latter one because it's unambiguous. In the first example the string `did not find system fact` repeats in both execution paths which requires additional effort to look for additional strings. Unless I'm mistaken, these log function calls can get interleaved with other logs which makes searching harder when reading logs.

kopiczko:
> Unless I'm mistaken, these log function calls can get interleaved with other logs which makes searching harder when reading logs.

Right now all CRs reconciliation is executed one by one. We have some plans to make it async, but anyway I think we most grep logs by object key anyway so I don't see this being a problem. Especially when we need to do so `canceling resource` logs anyway.

I still think it's the right thing to do, but I don't see ppl convinced. I leave that open for the next day to see opinions, and close if no consensus is met.

kopiczko:
I still think it makes sense, but I see you aren't enthusiastic. Closing.

PR #29, opened 2018-10-24
Adds todo guidelines

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/4463

### DIFF
--- a/go/todo.md
+++ b/go/todo.md
@@ -1,0 +1,16 @@
+# TODOs
+
+When writing a TODO in code, create a TODO issue
+(see https://github.com/giantswarm/giantswarm/issues/new?template=todo.md)
+and link it in the code.
+
+This helps ensure that TODOs don't die in the codebase,
+but are actioned by SIGs.
+
+e.g:
+```
+// TODO: Remove the foobar once the jank is lollobazzed.
+// 
+//	See https://github.com/giantswarm/giantswarm/issues/9001
+//
+```


go/todo.md
```diff
@@ -0,0 +1,14 @@
...
+and link it in the code.
+
+This helps ensure that TODOs don't die in the codebase,
+but are actioned by SIGs.
+
+e.g:
+```
+// TODO: Remove the foobar once the jank is lollobazzed.
+// See https://github.com/giantswarm/giantswarm/issues/9001
+```
```
xh3b4sd: Here comes the opinion train. I did this so far. 

```
// TODO foo bar. 
// 
//     https://github.com/issue/link
// 
```

This looks nicer in godoc and the link is easier to catch with the eyes. The 4 character indentation is displayed in a special way in the browser as well.

xh3b4sd (approved):
Go with what you think is right.

go/todo.md
kopiczko: +1 To be compatible with godoc (even if it isn't a strict godoc) it should should be actually a tab. See https://golang.org/src/text/template/doc.go for example.

go/todo.md
```diff
@@ -0,0 +1,14 @@
+# TODOs
+
+When writing a TODO in code, create a TODO issue
+(see https://github.com/giantswarm/giantswarm/issues/new?template=todo.md)
+and link it in the code.
+
+This helps ensure that TODOs don't die in the codebase,
+but are actioned by SIGs.
```
r7vme: -> it will die in github :D

r7vme (approved):

oponder (approved):

go/todo.md
JosephSalisbury: Done, with a tab. Bam!

PR #33, opened 2018-11-22
improves naming for custom resource

xh3b4sd:


MarcelMue (approved):
Fine with me.

go/unit_tests.md
```diff
@@ -43,12 +43,12 @@ Example function to be tested:
 ```go
 
 func ClusterID(obj interface{}) (string, error) {
-	customObject, ok := obj.(SpecificType)
+	customResource, ok := obj.(SpecificType)
```
kopiczko: How about `cr`?

kopiczko: I agree, but this is so commonly used that we all know about it and we can save some characters. Like with `ctx`.

PR #36, opened 2019-01-02
Add version bundle annotation

rossf7:
See https://github.com/giantswarm/chart-operator/pull/171#discussion_r244788534.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -6,6 +6,9 @@
   location. For now this should be in our `giantswarm/giantswarm` repository in
   which we crosslink all pages to create a reasonable documentation of all kinds
   of Kubernetes resources, their functionality and relationships.
+- `giantswarm.io/version-bundle` - for CRs exposed to customers we will use
+  this annotation instead of having it in the spec. As this is an implementation
+  detail for our operators.
 
 ## Common Labels (for control plane services)
 


rossf7:
I added sig-ux too because this came out of the App Catalog spec where we tried to make the new CRDs easier to understand for customers.

https://github.com/giantswarm/giantswarm/pull/4577

tomahawk28 (approved):
LGTM :)

kubernetes/annotations_and_labels.md
```diff
@@ -6,6 +6,9 @@
   location. For now this should be in our `giantswarm/giantswarm` repository in
   which we crosslink all pages to create a reasonable documentation of all kinds
   of Kubernetes resources, their functionality and relationships.
+- `giantswarm.io/version-bundle` - for CRs exposed to customers we will use
```
kopiczko: Let's please replace it with just `version`. It will make more sense with versioning of operators and in general. This is just a release version or operator version. `bundle` is just a legacy word. I also would express what it's actually is and it is going to replace `.spec.versionBundle` completely.

```suggestion
- `giantswarm.io/version` - for CRs exposed to customers we will use
```

kopiczko (approved):

kubernetes/annotations_and_labels.md
rossf7: Yes I see this as a stopgap solution until we move to `apiVersion`.

It keeps the new CRDs cleaner by moving the version to an annotation. Using `version-bundle` as the name matches the terminology we use currently.

puja108 (approved):

marians:
ack

PR #37, opened 2019-01-16
kubernetes: structure labels page

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5084.

This became messy over time. I'd like to structure it a bit for upcoming changes for VOO and app catalog.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -1,5 +1,7 @@
 # Kubernetes Annotations and Labels
 
+This page defines common annotations and labels we set in Kubernetes objects.
+
 ## Common Annotations
 
 - `giantswarm.io/docs` - value should be an URL to a common documentation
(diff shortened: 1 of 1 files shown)


kubernetes/annotations_and_labels.md
```diff
@@ -10,47 +12,80 @@
...
   defined in github.com/giantswarm/certs repo. This is used in certificate
   Secrets and CertConfigs.
+- `giantswarm.io/cluster` - value should contain tenant cluster ID which this
+  object is part of. E.g. `giantswarm.io/cluster=eggs2`.
 - `giantswarm.io/managed-by` - value should contain repository name of the
   operator managing the object. E.g. `giantswarm.io/managed-by=kvm-operator`.
 - `giantswarm.io/organization` - value should contain tenant cluster's
   organization ID as displayed in the front-end
-  `giantswarm.io/organization=track-hunter`.
+  `giantswarm.io/organization=adidas`.
```
kopiczko: This is the only real change I made. I replaced that with adidas because now we have public customers. Other stuff is just moved around untouched.

kubernetes/annotations_and_labels.md
```diff
@@ -10,47 +12,80 @@
...
+#### Tenant Cluster
+
+- `giantswarm.io/managed-by`
+
+### Labels Set In Secrets
+
+#### Control Plane
+
+- `giantswarm.io/cluster`
+- `giantswarm.io/certificate` - Only when `giantswarm.io/randomkey` is not set.
```
kopiczko: OK this `Only when` comments are extra.

xh3b4sd (approved):
Makes sense to me.

rossf7 (approved):
LGTM just a suggestion to change the org example

kubernetes/annotations_and_labels.md
kopiczko: Ah yes this is actually created by our customer. Yes I think `dev` is good example.

PR #39, opened 2019-01-17
kubernetes: update versioning of operators related labels

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5084.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -8,9 +8,6 @@ This page defines common annotations and labels we set in Kubernetes objects.
   location. For now this should be in our `giantswarm/giantswarm` repository in
   which we crosslink all pages to create a reasonable documentation of all kinds
   of Kubernetes resources, their functionality and relationships.
-- `giantswarm.io/version-bundle` - for CRs exposed to customers we will use
-  this annotation instead of having it in the spec. As this is an implementation
-  detail for our operators.
 
 ## Common Labels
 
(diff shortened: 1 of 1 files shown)


kubernetes/annotations_and_labels.md
```diff
@@ -34,13 +31,29 @@ This page defines common annotations and labels we set in Kubernetes objects.
...
-- `release.giantswarm.io/version` - value should be the used Giant Swarm release
-  version, e.g. `2.3.0`.
-- `OPERATOR.giantswarm.io/version` - value should be the used operator version
-  bundle version, e.g. `kvm-operator.giantswarm.io/version=1.0.0`.
+- `release.giantswarm.io/version` - value should be Giant Swarm release
+  version, e.g. `release.giantswarm.io/version=2.3.0`.
+- `OPERATOR.giantswarm.io/version` - value should be an operator version, e.g.
+  `kvm-operator.giantswarm.io/version=1.0.0`. This is not the same as release
+  version. This value can be the same in multiple releases. This label is used
+  by the operator to recognize which object it should reconcile.
```
kopiczko: Labels ending with `/version` are used for nodes. I completely forget how they are used. Let's make sure there is no interference or find out strategy to solve it.

xh3b4sd (approved):

TheoBrigitte (approved):

kubernetes/annotations_and_labels.md
```diff
@@ -34,13 +31,29 @@ This page defines common annotations and labels we set in Kubernetes objects.
...
 - `giantswarm.io/provider` - value should be the installation's provider, e.g.
   `kvm`, `aws`, or `azure`.
 
+### Labels Set In Custom Resources
+
+#### Control Plane
+
+- `giantswarm.io/managed-by`
+- `OPERATOR.giantswarm.io/version`
+- `release.giantswarm.io/version`
```
TheoBrigitte: I do not see why we need to have the release version set in CR. AFAIUI the CR only require an operator version to be reconciled and you can get the release version it belongs to from the release.

When an operator version belongs to multiple release version, which version should we put there?
Who is reconciling this label ?

kopiczko: You are right. No component from the VOO story is really using that. But still I think it's useful general information. Also for cross checking and monitoring.

TheoBrigitte: I agree it is nice to have this information, but I see it as being more complex than valuable. It means we would have to reconcile the release and operator version given by release CRs to every CR labels, while the point of truth still remain the release CR.
IMO I would rather keep it simple and not have `release.giantswarm.io/version` on CR.

kopiczko: So you have a request saying "upgrade to cluster `eggs2` release to `v1.2.3`" which means that we have to (among other things) change `eggs2` CR `OPERATOR.giantswarm.io/version` label value to the version of OPERATOR component set in that release. Where is the complexity of also changing `release.giantswarm.io/version` value to `v1.2.3`?

TheoBrigitte: Updating a label is not complex, but my point is about the higher level state replication complexity: the information about "operator version belong to which release(s)" is written in releases. When we want to replicate this info into all CRs, stuff can go wrong, so information written to CRs labels might not be correct. Therefore we should rather trust the information we have in releases than CRs. So I would not invest time in `release.giantswarm.io/version` which does not feel like the source of truth to me.

Proceed as you see a fit, just expressed my concern.

kopiczko: Ack. I still believe it's a zero effort. This isn't an information we rely on but I see it useful when debugging things. I wish I had that before so many times. That doesn't mean you don't have to cross check things if you don't have trust to the system yet. But I see it being a big time saver.

PR #41, opened 2019-01-29
Start a Javascript Section

oponder:
Since more people will be writing javascript, it's useful to agree on
how we do things.

So far it has been quite loose when compared to the standard we have so far
for our Go code. Aim is to improve that over time, and record our decisions
here.

### DIFF
--- a/javascript/general_style_guide.md
+++ b/javascript/general_style_guide.md
@@ -1,0 +1,13 @@
+# Javascript Style Guide
+
+Our Javascript Code Style is determined by `prettier`. Please make sure files have
+passed through `prettier` before committing them to the repo.
+
+A CI step will enforce that this has happened, failing the CI if it detects that
+`prettier` would make any changes.
+
+We use the following config params:
+- `--jsx-single-quote`
+- `--single-quote`
+- `--trailing-comma es5`
+
--- a/javascript/imports.md
+++ b/javascript/imports.md
@@ -10,6 +10,7 @@ Example:
 import 'babel-polyfill';
 import adminLogin from './auth/admin';
 import configureStore from '../stores/configureStore';
+import { ConnectedRouter } from 'connected-react-router';
 import forgot_password_index from './forgot_password/index';
 import forgot_password_set_password from './forgot_password/set_password';
 import history from '../stores/history';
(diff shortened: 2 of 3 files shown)


javascript/general_style_guide.md
```diff
@@ -0,0 +1,13 @@
+# Javascript Style Guide
+
+Our Javascript Code Style is determined by `prettier`. Please make sure files have
+passed through `prettier` before committing them to the repo.
```
tuommaki: I normally align this with `make run-prettier`, but would it make sense to write here simple instructions how to set it up to VS.Code which I think many use for front-end stuff? I noticed there are tons of plugins for this and I'm not sure if the default style in JS world is the same like `gofmt` in Go world. Or does this make sense at all?

javascript/imports.md
```diff
@@ -0,0 +1,32 @@
+# Imports
+
+Imports should be sorted alphabetically. That's all.
```
tuommaki: Is there a tool for this similar to `goimports` or is this something that developer must manually take care of?

tuommaki (approved):
LGTM. Couple question on matters I was curious.

javascript/general_style_guide.md
oponder: Javascript is no where near as standardised when it comes to tooling.

There is no 'standard' type of thing, though lots of people have been accepting 'prettier' as the formatter these days. The way to invoke it (whether by `make` or some other task runner) is not really standardised.

Also lots of different editors. I for example use Sublime still... though I might switch to VS.Code.

javascript/imports.md
oponder: For now something the developer must take care of manually. But there are ways to get this into the linter so that an error shows up, and then possibly also into something that does it automatically for us as part of a code style step, that we then add to `make` along with `prettier`.

But for now, keeping it simple.

javascript/general_style_guide.md
xh3b4sd: I tried for ten minutes and gave up to set it up in Atom.

xh3b4sd (approved):
I like the effort to emphazise standards a lot and when people are kind of happy with it I would say we go from here. Nice!

marians:
ACK

PR #43, opened 2019-01-31
go: add action key to logs

kopiczko:
This is a proposal for having an action key in the logs. Why do I want
to introduce it:

- Filtering logs by action.
- Recognize where the action starts and ends w/o looking at logs around.
- Values of `message` key in our logs got so structured that it's
  difficult to squeeze some additional information into them.
- Not all APIs are the equal. We need some more freedom to express what
  is going on.
- Reason logs didn't really kick in. It's hard to see their context.
- In the future I'd like us to be able to pass the action to the error.
  So this is visible when error is printed outside.

### DIFF
--- a/go/log_messages.md
+++ b/go/log_messages.md
@@ -37,13 +37,13 @@ We use `*context` packages for control flow of the reconciliation loop. Here are
 log messages associated with specific actions.
 
 ```go
-r.logger.LogCtx(ctx, "level", "debug", "action": "...", "message", "keeping finalizers")
+r.logger.LogCtx(ctx, "level", "debug", "message", "keeping finalizers")
 finalizerskeptcontext.SetKept(ctx)
 
-r.logger.LogCtx(ctx, "level", "debug", "action": "...", "message", "canceling reconciliation")
+r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
 reconciliationcanceledcontext.SetCanceled(ctx)
 
-r.logger.LogCtx(ctx, "level", "debug", "action": "...", "message", "canceling resource")
+r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
 resourcecanceledcontext.SetCanceled(ctx)
 ```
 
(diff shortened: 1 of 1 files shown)


xh3b4sd:
My question would be why we need a new logging key. I would suggest to use the log level `info` and put your "action" into the `message` key we already have. Then the structure would stay the same and we could simply filter for log levels.

tuommaki (commented):
I'm not fully sure if this is useful or not. I'm not straight out against this, but I'd really like to get this somehow implemented by automation in library API or something so that one would never need to write these manually. Even though these are only log lines, these feel similar to manual transaction handling or complex mutex locking which is proven to be error prone and when we start relying on this and forget some edge case, the logs are more harmful than helpful. I don't have good suggestion how to yield this, but just stating my thoughts here.

kopiczko:
> My question would be why we need a new logging key. I would suggest to use the log level info and put your "action" into the message key we already have

I can't imagine how this could look like. How could be the action log filtered? I'd like to be able to filter out levels within an action as well.

kopiczko:
> I'd really like to get this somehow implemented by automation in library API or something so that one would never need to write these manually

How do you imagine it to be less manual? There is only `actionLogger` setup step. Even with some library/instrumentation you'll have to set the context somewhere. With instrumentation you'll most likely have `caller` field set incorrectly or it won't be really helpful.

>  Even though these are only log lines, these feel similar to manual transaction handling or complex mutex locking which is proven to be error prone and when we start relying on this and forget some edge case, the logs are more harmful than helpful.

If we miss `start` or `end` log that's not a problem. We can just add it later on. We forget about some logs all the time. That's alright IMO. But when they are there they can make it easier to debug the thing. When you try to debug something and you see missing logs you add them to make debugging possible next time when a problem occurs. That doesn't affect running code. We should never rely on logs in other systems. This is like relying on monitoring data. Logs are only diagnostic/debugging information.

kopiczko:
> My question would be why we need a new logging key.

I tried to explain why I want to introduce it in the description.

kopiczko:
I think that rules we have for the `message` key so far as so fuzzy that they are difficult to apply. I sometimes struggle to make logs fit them and I have enough suggesting fixes to logs in reviews. I think what is proposed here is rather straight forward and well described. Also gives a lot of freedom also when it comes to printing additional information. This change here is not proposed to increase complexity of rules we have, but rather to replace them with something easier to follow.

MarcelMue:
I agree with Pawel in general here, I think we are currently a bit fuzzy here. My idea for a approach was different so far. I will have a think about this, but I am not against the proposal here.

xh3b4sd:
I do not think the `action` key proposal is an improvement. It is an additional concept with additional constraints. `"message", "start"` is not an improvement. "start" is not a message. The proposal does not make sense to me.

go/log_messages.md
```diff
@@ -76,16 +86,19 @@ The fact gathering statement expresses what happens and in case the action fails
...
 ```go
-s.logger.Log("level", "debug", "message", "collecting metrics")
-
-for _, c := range s.collectors {
-	err := c.Collect(ch)
-	if err != nil {
-		s.logger.LogCtx(ctx, "level", "error", "message", "failed to collect metrics", "stack", fmt.Sprintf("%#v", microerror.Mask(err)))
+{
+	actionLogger := r.loggger.With("action", "collecting metrics")
+	s.logger.LogCtx(ctx, "level", "debug", "message", "start")
```
TheoBrigitte: I think you should use actionLogger instead of s.logger, in orde to output the "action" key on start and end messages.

go/log_messages.md
```diff
@@ -76,16 +86,19 @@ The fact gathering statement expresses what happens and in case the action fails
...
+	for _, c := range s.collectors {
+		err := c.Collect(ch)
+		if err != nil {
+			actionLogger.LogCtx(ctx, "level", "error", "message", fmt.Sprintf("collector %#q failed", c.name), "stack", fmt.Sprintf("%#v", microerror.Mask(err)))
+		}
 	}
-}
 
-s.logger.LogCtx(ctx, "level", "debug", "message", "collected metrics")
+	s.logger.LogCtx(ctx, "level", "debug", "message", "end")
```
TheoBrigitte: same here

TheoBrigitte (approved):
I like it, more info in logs!

kopiczko:
Didn't get enough traction. Instead I'll try to collect problems I see with the logs we have so far and create an issue.

PR #44, opened 2019-02-12
add glossary proposal

xh3b4sd:
I think it would make sense to start writing down a little bit more about our terminologies. So we can be sure to talk about the same things. Clarifying the abbreviations also helps when writing in comments and the like.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,19 @@
 
 (Not only) coding guidelines and best practices hub.
 
+
+
 ## Azure
 
 - [Naming Resources](azure/naming_resources.md)
+
+
+
+## Glossary
+
+- [Control Plane](glossary/control_plane.md)
+- [Installation](glossary/installation.md)
+- [Tenant Cluster](glossary/tenant_cluster.md)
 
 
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -2,12 +2,22 @@
...
 
 - [Naming Resources](azure/naming_resources.md)
 
 
 
+## Glossary
+
+- [Control Plane](glossary/control_plane.md)
+- [Installation](glossary/installation.md)
+- [Tenant Cluster](glossary/tenant_cluster.md)
```
kopiczko: I feel like the whole glossary should be rather a single page. You can maybe put it under `General` section. I can see over time becoming really long list of single sentence pages.

kopiczko (approved):

README.md
xh3b4sd: I found it quite cool to see the separate files and the list in the readme. I don't think this list will become very long. We are not going to have more than 10 GS specific terms to talk about I guess. Or, I hope. :D

rossf7: I'd also be fine with a single page. But no blocker if you'd prefer them to be separate.

rossf7 (approved):
LGTM

README.md
tuommaki: I'd also prefer single page for easier navigation (or more like lack of unneeded navigation), but no blocker for me as well.

tuommaki (approved):

README.md
xh3b4sd: When the majority thinks a single page is better, let's do this when the list gets longer, or feel free to provide a PR. Thank you for your feedback.

PR #46, opened 2019-02-27
add suggestion to improve unit tests

xh3b4sd:


go/unit_tests.md
```diff
@@ -150,6 +150,7 @@ func Test_ClusterID(t *testing.T) {
 - one can run single test cases in table-driven tests
 - single test cases identifiers in test run are always in sync
 - useful diff comparison
+- run `go test ./... -run Test_Foo/5` to only execute a single test
```
kopiczko: In general this is easy to make error where `case N` and `Test_Foo/N` have different `N`, but I really like this oneliner.

kopiczko (approved):

go/unit_tests.md
xh3b4sd: Totally, but it is super handy.

PR #50, opened 2019-03-28
fix typo in golang code example

wind0r:
looks like return also got a extra `c` which should be removed ;)


JosephSalisbury:
Thanks for the help! <3

PR #51, opened 2019-04-10
go: add page about handling panic and os.Exit

kopiczko:


go/panic.md
```diff
@@ -0,0 +1,52 @@
...
+```go
+func main() {
+	err := mainE()
+	if err != nil {
+		panic(fmt.Sprintf("%#v", err))
+	}
+}
+
+// mainE is a helper function allowing main to handle os.Exit in a single place.
+func mainE() error {
```
kopiczko: We have different approaches on naming. I took inspiration from cobra. This is sexy and straight to the point.

go/panic.md
```diff
@@ -0,0 +1,53 @@
...
+can not return an error.
+
+The handling should look like this:
+
+```go
+func main() {
+	err := mainE()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%#v", err)
+                os.Exit(1)
```
MarcelMue: Whitespace looks weird here

go/panic.md
MarcelMue: We used `mainWithError` so far, right?
I would not get the meaning of `mainE` right away to be honest.

MarcelMue (approved):
Fine with me over all.

go/panic.md
kopiczko: We used `mainError`/`mainWithError`.

go/panic.md
kopiczko: I don't see what you mean. Can you add a suggestion?

MarcelMue: I can't really do that with the github UI.
For me `os.Exit(1)` should be on the same indentation of `fmt.Fprintf(os.Stderr, "%#v", err)` but is way more indented.

kopiczko:
```suggestion
		os.Exit(1)
```

kopiczko: Now I get this. There were spaces instead of tabs. It looked ok in my browser. Thanks.

tuommaki (approved):

go/panic.md
rossf7: I prefer `mainWithError` for me it's clearer and we only use it once per project.

rossf7 (approved):
LGTM

go/panic.md
kopiczko: My point here is `main` isn't the only case. I just gave the most prominent example. I still prefer short form because it introduces less noise - especially for longer function names. But I'll follow the quorum.

kopiczko:
I go with `E` suffix. I don't perceive this as an annoyance from the comments and I saw longer function names with 'WithError` suffix and this them harder to read (at least for me).

PR #54, opened 2019-06-12
Add dockerfile coding conventions

rossf7:
Follow up to https://github.com/giantswarm/giantswarm/pull/6191 as here is a better fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,18 @@
 
 
 
+## Docker
+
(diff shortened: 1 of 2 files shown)


kopiczko (approved):
I also think this is better place for that. Hopefully we can keep that up to date.

MarcelMue (approved):
fine with me :heart:

TheoBrigitte (approved):
Beautiful

PR #55, opened 2019-06-26
Use imperative form for changelogs

rossf7:
The docs are outdated here because I believe the consensus is now that we should use the imperative form rather than past tense.

e.g. `Update to 1.5.0.` and not `Updated to 1.5.0.`

If you disagree please raise your concerns here.


MarcelMue (approved):
I was living in the past all along :D 

:+1:

xh3b4sd (approved):

rossf7 (commented):
This is `fmt` so fishing for at least one more approval.

kopiczko (approved):
Thanks <3

PR #60, opened 2019-07-04
suggest improvements for cloud formation stack naming scheme

xh3b4sd:
In scope of the Node Pools story we rename a couple of things. So for instance the the CF stacks. Tuomas brought up how the new CF stack should be called and it would be good to have a reasonable naming scheme somehow. Right now we have this. 

```
CPI - Control Plane Initializer
CPF - Control Plane Finalizer
```

```
TCCP - Tenant Cluster Control Plane
TCDP - Tenant Cluster Data Plane
```

How should this be named? In this PR I made a suggestion on which your feedback is required. See also https://github.com/giantswarm/aws-operator/pull/1725#discussion_r300400240.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 - [Control Plane](glossary/control_plane.md)
 - [Installation](glossary/installation.md)
 - [Tenant Cluster Control Plane](glossary/tenant_cluster_control_plane.md)
-- [Tenant Cluster Data Plane](glossary/tenant_cluster_data_plane.md)
+- [Tenant Cluster Node Pool](glossary/tenant_cluster_node_pool.md)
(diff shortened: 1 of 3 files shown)


MarcelMue (approved):
Fine with me

xh3b4sd:
More please. I want 3 or 4 approvals on this bad boy.

kopiczko (approved):

xh3b4sd:
@giantswarm/sig-operator please review and approve when you agree or discuss at least. I would like to have at least two more approvals here.

xh3b4sd:
I think everyone should have read their Emails after 24 hours. When nobody wants to express their opinion further, I go with this now.

PR #66, opened 2019-07-26
Document the catalog-type label.

oponder:
This documents the `application.giantswram.io/catalog-type` label which all `AppCatalog` CR's are expected to have from now on.

ping @JosephSalisbury and anyone else creating Control Plane catalogs, please include this new label and set it to `internal`, then Happa will know not to show it in the UI. (To be implemented)

I think in general I will make it so that Happa does not show any catalog when this value is empty. but we can discuss that in a soon to come Happa PR.

Related to: https://github.com/giantswarm/giantswarm/issues/6616


sslavic (approved):
LGTM

rossf7 (approved):
LGTM

PR #75, opened 2019-10-09
Add guidelines for templating images in helm charts

rossf7:
Prepares us for using App Catalog and reducing some of the image pulling pain we see in AWS China.

### DIFF
--- a/helm/image_templating.md
+++ b/helm/image_templating.md
@@ -1,0 +1,17 @@
+# Helm Chart Image Templating
+
+## Templating
+
+When templating container images in charts `values.yaml` should be structured
+like this. This allows us to override `image.registry` in some control planes
+e.g. AWS China.
+
+
+```yaml
+image:
+  registry: quay.io
+  name: giantswarm/kube-state-metrics
+  tag: v1.8.0
+```
+
+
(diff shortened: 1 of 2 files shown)


helm/image_templating.md
```diff
@@ -0,0 +1,17 @@
...
+## Templating
+
+When templating container images in charts `values.yaml` should be structured
+like this. This allows us to override `image.registry` in some control planes
+e.g. AWS China.
+
+
+```yaml
+image:
+  registry: quay.io
```
tomahawk28: Not related to this change but most of the charts are simply `quay.io` images, can we may set up the default value for that?

sslavic (approved):
LGTM

helm/image_templating.md
rossf7: Yes we use `quay.io` everywhere except AWS China where we use `registry-intl.cn-shanghai.aliyuncs.com`.

But at some point we may want to run our own registry like Harbor and for some onprem control planes as well as China.

tomahawk28 (approved):

rossf7:
Fishing for 1 more approval since this is `fmt`.

tfussell (approved):
Makes sense to me!

PR #78, opened 2019-10-29
Add screenshot to intellij idea config for imports

fiunchinho:


MarcelMue:
Change in general is okay, but can we put the image in some sub folder?
like `images` or something?

tuommaki:
Would it make sense to also group current project imports? Not sure how that behaves, but it sounds like what we desire.

fiunchinho:
Moved to subfolder and updated image with option "group current imports" checked.

PR #80, opened 2019-11-08
go: improve example for unit tests

kopiczko:
It should print the name only for failed tests (unless -v flag is used).


tuommaki (approved):

MarcelMue (approved):
LGTM

PR #83, opened 2019-11-26
JS strings

ograu:


javascript/general_style_guide.md
```diff
@@ -15,6 +15,18 @@ Please use always triple equal comparisions and get rid of a lot of bugs related
 "0" === 0 // false
 ```
 
+### Strings
+
+Use single quotes for strings: `'This is a regular string` 
+
+Use template strings for interpolation or multiline strings:
```
marians: Is "template strings" a term?

How about "backticks and template strings", to clarify the quoting?

javascript/general_style_guide.md
```diff
@@ -106,7 +117,7 @@ with its pros and cons. More info in the [React Docs](https://reactjs.org/docs/h
 
 ### Naming
 
-- Files: snake_case
+- Files: cameCase
```
marians:
```suggestion
- Files: camelCase
```

marians: Wait, so this means that when we have a component `ClusterDashboardItem` the fiel name should be `clusterDashboardItem.js` instead of `ClusterDashboardItem.js`?

ograu: I've have added more text, hope it makes sense now.

javascript/general_style_guide.md
ograu: Explanation in the template string example.

marians (approved):

ograu:
@giantswarm/sig-front-end  lets merge this?

oponder (approved):
LGTM! Thanks for writing it up.

oponder (approved):
Approving as sig-ux as well :P or did you want some specific UX feedback on something?

ograu:
XD no it's just I need another approval!  ping @axbarsan

javascript/general_style_guide.md
```diff
@@ -41,7 +54,6 @@ We use the following config params:
 - `--single-quote`
 - `--trailing-comma es5`
```
axbarsan: Do we still need to have these here? We already have the `prettier` configuration in our `package.json`

axbarsan (changes requested):

javascript/general_style_guide.md
ograu: Probably we don't need this. We can keep it here just for reference or get rid of this.

javascript/general_style_guide.md
```diff
@@ -106,9 +118,9 @@ with its pros and cons. More info in the [React Docs](https://reactjs.org/docs/h
 
 ### Naming
 
-- Files: snake_case
+- Regular JS Files (not components): camelCase.js
+- Component Files: PascalCase
```
oponder: how about the folder names? do we keep them underscore and lowercase?

ograu: I don't have any specific opinion on that, but since @ubergesundheit  is going to work on it, let's agree on something... ¿What about pascal case for folders containing components and lowercase for folders not containing components?

I like this way because the folder name gives you information of what it's inside.

javascript/general_style_guide.md
ograu: @oponder since you wrote this, ¿Should we remove this or not?

oponder: Yup, I'm ok with removing it.

javascript/general_style_guide.md
ubergesundheit: I guess either is fine with me but using camelCase for folders would be more in line with the filename convention. I like using PascalCase for folders containnig components.

ograu: I like it!

oponder: ok, I can get used to that!

oponder: oh and then the folder should always have a file in it with the same name, which will be the 'root' component right. I used to use `index.js` for that, but we want to avoid that right?


So if we have a folder: `ClusterView` then there should always be a file `ClusterView.js` in there too.

ograu: That would be ideal IMO

axbarsan (approved):

PR #85, opened 2019-12-05
Happa testing

ograu:
This is more or less what we agreed like two months ago, but as I said, this is nothing set in stone, so please feel free to suggest any changes or open any discussion regarding any of what is added in this PR.

ps: I like using the word manifesto, it makes me feel I'm writing super important stuff 😛 
Let's delete it if you find it too pendantic!

Thanks

### DIFF
--- a/javascript/testing.md
+++ b/javascript/testing.md
@@ -25,7 +25,7 @@ Use them just in components that we are not changing too frequently. Otherwise w
 We want our tests to be [as resilient as possible](https://twitter.com/dan_abramov/status/1151842775417151489). Hence, don’t test if a css class is present or a piece of state holds this or that value, test what is actually rendered. For example:
 
 - Test if a button is red instead of it has an `alert` class
-- Test if an element is in present in the DOM instead of testing a state value in a component
+- Test if an element is present in the DOM instead of testing a state value in a component
 
 ### Unit tests vs Integration tests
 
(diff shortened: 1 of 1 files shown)


javascript/testing.md
```diff
@@ -1,79 +1,44 @@
...
-})
-```
+Use them just in components that we are not changing too frequently. Otherwise we will have to update them every now and then.
+
+### Avoid implementation details
+
+We want our tests to be [as resilient as possible](https://twitter.com/dan_abramov/status/1151842775417151489). Hence, don’t test if a css class is present or a piece of state holds this or that value, test what is actually rendered. For example:
+
+- Test if a button is red instead of it has an `alert` class
+- Test if an element is in present in the DOM instead of testing a state value in a component
```
oponder:
```suggestion
- Test if an element is present in the DOM instead of testing a state value in a component
```

ograu:
Hey @ubergesundheit @axbarsan ¿Are you ok with this?

ubergesundheit:
LGTM. One question though: Are the test examples not valid any more or why are they removed?

ograu:
The tests were there because Marian asked examples. I removed there because tests in the codebase are better examples than those. Also because we are writing tests in a different way, eg: we don't use snapshot tests, we are not mocking API calls with `jest.mock`, etc.

oponder:
We need another approval here :) ping @giantswarm/sig-front-end are we good to go on these fmt changes?

ograu:
Hey @axbarsan Are you ok with this? I would like to merge it.
Thanks :-)

PR #89, opened 2020-01-09
go: don't encourage to wrap arbitrary errors

kopiczko:


go/errors.md
```diff
@@ -14,24 +14,18 @@ We use [microkit] to enrich **all** errors with stack traces.
...
 
-For errors returned by function call it is fine to just wrap them and return:
+While creating new error, then error from the same package (i.e. from error.go
+file) should be wrapped with `Maskf` with additional information. The
+additional information should be lowercased:
 
 ```go
-return microerror.Mask(err)
-```
+return microerror.Maskf(executionFailedError, "failed to request %#q", url)
```
tfussell: Why `%#q` instead of `%s`? It seems that this adds backquotes around the string. Is this for `luigi`?

xh3b4sd (approved):

go/errors.md
xh3b4sd: https://github.com/giantswarm/fmt/blob/master/go/log_messages.md#general-guidelines

kopiczko: No this is rather for die-hard JSON readers so they don't see `\"` when using `#q`. And we want formatted strings quoted one way or another.

kopiczko:
This is not controversial so I won't wait for all 3 approvals.

PR #91, opened 2020-01-17
Add annotation for restart purposes

tuommaki:
Add generic restart annotation that can be used e.g. to issue rolling restart
of DamonSets, Deployments and StatefulSets.

`kubectl` implements rolling restart of deployments by setting `kubectl.kubernetes.io/restartedAt` annotation (see [objectrestarter](https://github.com/kubernetes/kubectl/blob/release-1.16/pkg/polymorphichelpers/objectrestarter.go#L32)).

Initial use-case for this is in `azure-operator` to gracefully move workload from old nodes to new nodes during upgrade.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -8,6 +8,9 @@ This page defines common annotations and labels we set in Kubernetes objects.
   location. For now this should be in our `giantswarm/giantswarm` repository in
   which we crosslink all pages to create a reasonable documentation of all kinds
   of Kubernetes resources, their functionality and relationships.
+- `giantswarm.io/restart` - this annotation is meant to be used for DaemonSet,
+  Deployment and StatefulSet restarts. The value can be arbitrary string that
+  works for given reconciliation purposes.
 - `machine-deployment.giantswarm.io/subnet` - value contains a subnet CIDR that
   has been allocated for the given node pool / MachineDeployment object that
   has this annotation. E.g.


xh3b4sd (approved):
Works for me.

kubernetes/annotations_and_labels.md
```diff
@@ -8,6 +8,9 @@ This page defines common annotations and labels we set in Kubernetes objects.
   location. For now this should be in our `giantswarm/giantswarm` repository in
   which we crosslink all pages to create a reasonable documentation of all kinds
   of Kubernetes resources, their functionality and relationships.
+- `giantswarm.io/restart` - this annotation is meant to be used for DaemonSet,
+  Deployment and StatefulSet restarts. The value can be arbitrary string that
```
MarcelMue: Maybe I am a dummy here but I don't get the use case from this description - when would this annotation be added?

kopiczko: +1

kubernetes/annotations_and_labels.md
```diff
@@ -8,6 +8,9 @@ This page defines common annotations and labels we set in Kubernetes objects.
   location. For now this should be in our `giantswarm/giantswarm` repository in
   which we crosslink all pages to create a reasonable documentation of all kinds
   of Kubernetes resources, their functionality and relationships.
+- `giantswarm.io/restart` - this annotation is meant to be used for DaemonSet,
```
kopiczko: Here we should explain what actions this annotation imposes. Where it is used should be described in one of the sections at the bottom of the page.

kubernetes/annotations_and_labels.md
tuommaki: > when would this annotation be added?

When you want to programmatically issue a rolling restart of `Deployment`, `DaemonSet` or `StatefulSet`. I.e. when you want to implement `kubectl rollout restart` in some operator.

In my use-case I want to perform this in `azure-operator` during worker VMSS upgrade in order to gracefully move workload from old nodes to new nodes.

kubernetes/annotations_and_labels.md
tuommaki: I feel that the structure lower down in this document doesn't really work for this annotation. This is kind of generic and can be put on any object that is wanted to be updated. Do you have ideas how to change that structure or should this be just described here?

kubernetes/annotations_and_labels.md
kopiczko: Let's write that this should be added manually by the user in order to perform rolling restart.

kubernetes/annotations_and_labels.md
kopiczko: Now when you explained what this is supposed to do in the other comment I think it's fine. Let's improve the comment a bit to try to reduce the general confusion.

kubernetes/annotations_and_labels.md
```diff
@@ -8,6 +8,9 @@ This page defines common annotations and labels we set in Kubernetes objects.
...
+  Deployment and StatefulSet restarts. The value can be arbitrary string that
+  works for given reconciliation purposes.
```
kopiczko:
```suggestion
  works for given reconciliation purposes. It should be set manually by the user. After full restart the annotation is removed by restarting authority.
```

kopiczko (approved):

kubernetes/annotations_and_labels.md
tuommaki: No, in general this should not be used manually but programmatically (e.g. by `azure-operator` in my case). It also isn't supposed to be removed afterwards because that would trigger another rolling restart.

kopiczko: Alright. So this is just to change the DS/StatefulSet so to rolling update is triggered. With you now.

tuommaki:
After all discussion, I'm dropping this as my approach changed. Nevertheless, learning how the rolling restart works was still useful. Thank you and sorry guys :flushed:

PR #92, opened 2020-01-20
Suggest CircleCI best practice: step documentation

marians:
Dear SIG releng, with this PR I propose to make CircleCI steps more self-explanatory.

Here is an example of some nicely documented steps, as suggested in the text:

image

And here is an example of what happens if steps are not documented via the `name` field:

image

This is especially important to me, as these automations have to be understood by many people, but not frequently, so there are many people who don't know these steps by heart and would benefit from better readability and explanation, in case something goes wrong.

### DIFF
--- a/github/circleci/README.md
+++ b/github/circleci/README.md
@@ -1,0 +1,21 @@
+# CircleCI best practices
+
+## Making steps self-explanatory
+
+CircleCI offers a way to make a step in a CI config or org self-explanatory. We SHOULD always use this to make the config more self-explanatory.
+
+### Don't
+
+```yaml
+  steps:
+    - run: [ -z ${CIRCLE_TAG} ] && echo -n 'giantswarm-operations-platform-test-catalog' > .app_catalog_name || echo -n 'giantswarm-operations-platform-catalog' > .app_catalog_name
+```
+
+### Do
+
+```yaml
+  steps:
+    - run:
+      name: Set target catalog name in temp file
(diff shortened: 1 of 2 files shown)


rossf7 (approved):
Nice, good to have this documented. LGTM

github/circleci/README.md
```diff
@@ -0,0 +1,21 @@
+# CircleCI best practices
+
+## Making steps self-explanatory
+
+CircleCI offers a way to make a step in a CI config or org self-explanatory. We SHOULD always use this to make the config more self-explanatory.
```
oponder: Can we make this short and sweet with more direct language, and by removing a redundant sentence?

Like:

```suggestion
Always set the `name` of your steps to something that explains what it does.
```

kopiczko:
We already talked about it in the SIG. We want to do that. The downside is that the configuration would be like 3 times longer but I think it's ok price to pay. We also want to prefix steps with job names. Like `architect/push-to-docker: log in to docker`. I'm thinking about moving contents of this PR to `architect-orb` readme because this is where we should define majority of our configs. WDYT?

marians:
@kopiczko Not sure about moving, as it also applies to CircleCI config outside orbs. Rather no. Anyway, I'll merge and may the SIG go from here.

marians:
Well, won't merge, since I need 2 approvals :)

kopiczko (approved):

marians:
Wait, we need 3 approvals now? Pretty strict :)

oponder (approved):
LGTM!

PR #95, opened 2020-02-05
Redux selectors

ograu:


javascript/general_style_guide.md
```diff
@@ -352,14 +352,31 @@ In cases such as batched functions called for __refreshing data which is already
 
 This option will reset `cluster.nodePools` to an empty array in the store. This is the array that contain node pools IDs for a given cluster. We just do this the first time we fetch details for a cluster.
 
+### Selectors
 
-## Testing
+We access store from components using `mapStateToProps` (or `useSelector` hook alternatively), or selectors.
```
oponder:
```suggestion
We access the store from components using `mapStateToProps`, a `useSelector` hook, or selectors.
```

Does that make sense?

oponder: Actually, `selectors` would still be used in the mapStateToProps function right?

So it would be more like:


We can access the store from our components by using `mapStateToProps` or a `useSelector` hook.
In mapStateToProps we prefer that you use a `selector` function.

ograu: Yes, you are right, let me rewrite this please

ograu:
@ubergesundheit Are you ok with this?

ograu:
@ubergesundheit Are you ok with this?

ograu:
Thanks @ubergesundheit

PR #96, opened 2020-02-09
Add note excluding orbs for CircleCI recommendation

ubergesundheit:
I could not find the official CircleCI documentation on this, but it seems in orbs, CircleCI does not allow a `name` key in the `steps` element. (See [giantswarm/architect-orb#58](https://github.com/giantswarm/architect-orb/pull/58#pullrequestreview-354254513))

### DIFF
--- a/circleci/all-circleci-best-practices.md
+++ b/circleci/all-circleci-best-practices.md
@@ -2,7 +2,7 @@
 
 ## Making steps self-explanatory
 
-Always set the `name` of your steps to something that explains what it does.
+Always set the `name` of your steps to something that explains what it does. Note: This does not work for orbs as CircleCI does not allow a `name` key for elements in the `steps` array.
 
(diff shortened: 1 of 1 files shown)


ubergesundheit:
Sorry for not reading the note very well

circleci/all-circleci-best-practices.md
```diff
@@ -2,7 +2,7 @@
 
 ## Making steps self-explanatory
 
-Always set the `name` of your steps to something that explains what it does.
+Always set the `name` of your steps to something that explains what it does. Note: This does not work for orbs as CircleCI does not allow a `name` key for elements in the `steps` array.
```
kopiczko: Actually in orb I'd like to have a bit different convention including the job name as a prefix. It would be nice to be able to group steps of a single job.

PR #97, opened 2020-02-20
Explain waiting for nock to be done as part of your test.

oponder:
Explain a bit of what I found while trying to clean up request behaviour in the Organizations test.

### DIFF
--- a/javascript/testing.md
+++ b/javascript/testing.md
@@ -51,7 +51,7 @@ It:
 
 We should avoid using persisted nocks so that we get a good idea of the request
 behaviour of our components. You might run into trouble when trying to do this
-especially when combining it with nock.cleanAll(); in your afterEach block and nock.disableNetConnect(); before running the test suite.
+especially when combining it with `nock.cleanAll();` in your `afterEach` block and `nock.disableNetConnect();` before running the test suite.
 
 If you see errors after your test has finished running like:
 
--- a/javascript/testing.md
+++ b/javascript/testing.md
@@ -45,3 +45,47 @@ It:
 - Easy to read/understand
 - Catches bugs
 - Gives good coverage to effort ratio
+
+
+### Persisted nocks considered a antipattern
+
+We should avoid using persisted nocks so that we get a good idea of the request
+behaviour of our components. You might run into trouble when trying to do this
+especially when combining it with `nock.disableNetConnect();` and `nock.cleanAll();`
+in your `afterEach` block.
+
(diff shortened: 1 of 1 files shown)


javascript/testing.md
```diff
@@ -45,3 +45,47 @@ It:
 - Easy to read/understand
 - Catches bugs
 - Gives good coverage to effort ratio
+
+
+### Persisted nocks considered a antipattern
+
+We should avoid using persisted nocks so that we get a good idea of the request
+behaviour of our components. You might run into trouble when trying to do this
+especially when combining it with `nock.disableNetConnect();` and `nock.cleanAll();`
```
axbarsan: Not sure about the wording here, because it sounds like we're placing `nock.disableNetConnect();` in the `afterEach` block, which wouldn't really help.

Maybe reversing them would work?

"when combining it with `nock.cleanAll(); in your `afterEach` block and `nock.disableNetConnect();` before running the test suite"

ograu:
Thanks Oliver.

In order to use this effectively, we will have to refactor all tests to not use `persist()`, rigth?

ograu:
I am thinking that one problem with this approach is that it could increase the amount of time each test will need to finish... If this is true, it can be a huge problem...

oponder:
> In order to use this effectively, we will have to refactor all tests to not use persist()

Right, but I don't think we need to force that everywhere yet.

>  It could increase the amount of time each test will need to finish.

I don't think it'll have a very huge impact. but let's keep an eye on it.

The best thing that can come out of it is that we discover Happa makes more requests than necessary and we can fix that in a component to stop the test from waiting and so make the test faster again.

javascript/testing.md
oponder: Thanks! done!

javascript/testing.md
```diff
@@ -51,8 +51,7 @@ It:
 
 We should avoid using persisted nocks so that we get a good idea of the request
 behaviour of our components. You might run into trouble when trying to do this
-especially when combining it with `nock.disableNetConnect();` and `nock.cleanAll();`
-in your `afterEach` block.
+especially when combining it with nock.cleanAll(); in your afterEach block and nock.disableNetConnect(); before running the test suite.
```
axbarsan:
```suggestion
especially when combining it with `nock.cleanAll();` in your `afterEach` block and `nock.disableNetConnect();` before running the test suite.
```

PR #102, opened 2020-03-18
kubernetes: Integrate k8s common labels

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/8632

Include common labels advocated in Kubernetes docs, there's some
overlap with our internal ones and after discussing this the consensus
was that it would be better to avoid that and standardise on the well
defined set from Kubernetes/Helm upstream.

Those labels allow us to describe the source and instance of an app so
it's easy to see which installation an object is part of.

Inline description of the main labels specifying what are the expected
values and that they apply to all objects composing an app/operator
they refer to.

Also add this to description of related custom `.../branch` and
`.../commit` labels.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -56,18 +56,15 @@ This page defines common annotations and labels we set in Kubernetes objects.
   This together with `app.kubernetes.io/name` should be used as replicas
   selector, e.g.
   `app.kubernetes.io/name=kvm-operator,app.kubernetes.io/instance=kvm-operator-1.0.0`.
-- `app.kubernetes.io/version` and other common labels (see
-  [here][helm-labels] and [here][k8s-common-labels]) - are informational,
-  applied to all objects in a chart that installs an app or operator and so
-  they indicate the source and instance of that app/operator; value of
-  `.../version` should be the app/operator version as defined in `project.go`
-  and matching `appVersion` in `Chart.yaml`, e.g.
-  `app.kubernetes.io/version=1.0.0`.
-- `app.giantswarm.io/branch` - (informational) branch from which the instance
-  of the app/operator that this object is part of was built from.
-- `app.giantswarm.io/commit` - (informational) ID (git SHA) of the commit from
-  which the instance of the app/operator that this object is part of was built
-  from.
+- informational labels, applied to all objects in a chart that installs an app
+  or operator and so they indicate its source and instance:
+  - `app.kubernetes.io/version` - value should be the app/operator version as
+    defined in `project.go` and matching `appVersion` in `Chart.yaml`, e.g.
+    `app.kubernetes.io/version=1.0.0`.
+  - `app.giantswarm.io/branch` - branch from which the instance of the
+    app/operator that this object is part of was built from.
+  - `app.giantswarm.io/commit` - ID (git SHA) of the commit from which the
+    instance of the app/operator that this object is part of was built from.
 - `giantswarm.io/provider` - value should be the installation's provider, e.g.
   `kvm`, `aws`, or `azure`.
 
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -39,16 +39,16 @@ This page defines common annotations and labels we set in Kubernetes objects.
 - `release.giantswarm.io/version` - value should be Giant Swarm release
   version, e.g. `release.giantswarm.io/version=2.3.0`.
 - `OPERATOR.giantswarm.io/version` - value should be the version of the
-  operator reconciling the object, as defined in `project.go` and matching
+  given operator reconciling the object, as defined in its `project.go` and
   `appVersion` in `Chart.yaml`, e.g.
   `kvm-operator.giantswarm.io/version=1.0.0`. It is used by the given operator
   to recognize which object it should reconcile (i.e. to only reconcile objects
-  matching its own version). When set on Node objects it is used to set that
-  information in the status with the statusresource. This is different from
-  release version and can be the same in multiple releases. Its value may be
-  equal to that of `app.kubernetes.io/version` but it has a different purpose
-  and since there could be multiple operators reconciling one object there
-  could be multiple per-operator labels on one object.
+  matching its own version, which is exposed as `app.kubernetes.io/version` on
+  the operator's own component resources like `Deployment`). When set on Node
+  objects it is used to set that information in the status with the
+  statusresource. This is different from release version and can be the same in
+  multiple releases. Since there could be multiple operators reconciling one
+  object there could be multiple per-operator labels on one object.
 - `helm.sh/chart` - value should contain a chart name and version, see
   [Helm chart labels best practices][helm-labels]
 - `app.kubernetes.io/name` - the name of the application; should be applied to
@@ -63,7 +63,12 @@ This page defines common annotations and labels we set in Kubernetes objects.
   or operator and so they indicate its source and instance:
   - `app.kubernetes.io/version` - value should be the app/operator version as
     defined in `project.go` and matching `appVersion` in `Chart.yaml`, e.g.
-    `app.kubernetes.io/version=1.0.0`.
+    `app.kubernetes.io/version=1.0.0`. In case of our operators there should be
+    only a single instance with a given version running on a cluster as this
+    also indicates this operator reconciles objects labeled with
+    `OPERATOR.giantswarm.io/version` with the same value, and there should be
+    only one **version** of an operator reconciling a single object. This is
+    enforced by GateKeeper on our clusters.
   - `app.giantswarm.io/branch` - branch from which the instance of the
     app/operator that this object is part of was built from.
   - `app.giantswarm.io/commit` - ID (git SHA) of the commit from which the
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -254,7 +254,7 @@ with labels:
 <details>
 <summary>EXAMPLE</summary>
 <p>
-For example, given snippet from `templates/deployment.yaml` in _aws-operator_:
+For example, this snippet from <code>templates/deployment.yaml</code> in <em>aws-operator</em>:
 
 ``` yaml
 apiVersion: apps/v1
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -182,13 +182,13 @@ querying by shared tooling.
 Labels related to a higher-level virtual concept of an _app_, i.e. a bunch of
 components usually installed by one or more Helm charts.
 
-- `helm.sh/chart`
 - `app.kubernetes.io/name`
 - `app.kubernetes.io/instance`
-- `app.kubernetes.io/version`
-- `app.kubernetes.io/managed-by`
 - `app.giantswarm.io/branch`
 - `app.giantswarm.io/commit`
+- `app.kubernetes.io/managed-by`
+- `app.kubernetes.io/version`
+- `helm.sh/chart`
 
 To set those labels without having to repeat their definition in multiple
 places we use a template helper based on that in [chart template][helm-def-tpl]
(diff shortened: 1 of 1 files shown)


kubernetes/annotations_and_labels.md
```diff
@@ -51,16 +47,38 @@ This page defines common annotations and labels we set in Kubernetes objects.
...
   different purpose and since there could be multiple operators reconciling one
   object there could be multiple per-operator labels on one object.
-- `app.giantswarm.io/branch` - (informational) branch from which this
-  instance of the app/operator was built from.
+- `helm.sh/chart` - value should contain a chart name and version, see
+  [Helm chart labels best practices][helm-labels]
+- `app.kubernetes.io/name` - the name of the application; should be applied to
+  every Kubernetes resource associated with that application, i.e. all objects
+  in a chart that installs it.
+- `app.kubernetes.io/instance` - value should be set to `{{ .Release.Name }}`
```
kopiczko: It's somehow weird that we say what should be set here but not in the other labels.

kubernetes/annotations_and_labels.md
```diff
@@ -51,16 +47,38 @@ This page defines common annotations and labels we set in Kubernetes objects.
...
+  and is meant for differentiating between instances of the same application.
+  This together with `app.kubernetes.io/name` should be used as replicas
+  selector, e.g.
+  `app.kubernetes.io/name=kvm-operator,app.kubernetes.io/instance=kvm-operator-1.0.0`.
+- `app.kubernetes.io/version` and other common labels (see
+  [here][helm-labels] and [here][k8s-common-labels]) - are informational,
+  applied to all objects in a chart that installs an app or operator and so
+  they indicate the source and instance of that app/operator; value of
+  `.../version` should be the app/operator version as defined in `project.go`
```
kopiczko: This `.../version` shortcut is confusing IMO.

kubernetes/annotations_and_labels.md
```diff
@@ -51,16 +47,38 @@ This page defines common annotations and labels we set in Kubernetes objects.
...
+- `app.kubernetes.io/name` - the name of the application; should be applied to
+  every Kubernetes resource associated with that application, i.e. all objects
+  in a chart that installs it.
+- `app.kubernetes.io/instance` - value should be set to `{{ .Release.Name }}`
+  and is meant for differentiating between instances of the same application.
+  This together with `app.kubernetes.io/name` should be used as replicas
+  selector, e.g.
+  `app.kubernetes.io/name=kvm-operator,app.kubernetes.io/instance=kvm-operator-1.0.0`.
+- `app.kubernetes.io/version` and other common labels (see
+  [here][helm-labels] and [here][k8s-common-labels]) - are informational,
```
kopiczko: I would skip the other common labels part. This page is about labels we use and how we do this. If we want to use any of the common upstream labels it should be listed here.

kubernetes/annotations_and_labels.md
```diff
@@ -51,16 +47,38 @@ This page defines common annotations and labels we set in Kubernetes objects.
   value may be equal to that of `app.kubernetes.io/version` but it has a
   different purpose and since there could be multiple operators reconciling one
   object there could be multiple per-operator labels on one object.
-- `app.giantswarm.io/branch` - (informational) branch from which this
-  instance of the app/operator was built from.
+- `helm.sh/chart` - value should contain a chart name and version, see
+  [Helm chart labels best practices][helm-labels]
+- `app.kubernetes.io/name` - the name of the application; should be applied to
+  every Kubernetes resource associated with that application, i.e. all objects
+  in a chart that installs it.
```
kopiczko: It would be cool to add some example here.

kubernetes/annotations_and_labels.md
```diff
@@ -161,6 +179,20 @@ This page defines common annotations and labels we set in Kubernetes objects.
 
 #### Tenant Cluster
 
+### App Labels
```
kopiczko: Someone decided to skip one level in the following paragraphs so I think you want to do `##`.

```suggestion
## Chart Templates Labels
```

kubernetes/annotations_and_labels.md
mig4: Well, the Helm docs say `"This should be the {{ .Release.Name }}. "` so I thought it'd be best to include it here.

kubernetes/annotations_and_labels.md
mig4: Fair enough, I'll get rid of them.

kubernetes/annotations_and_labels.md
mig4: Sure, makes sense.

kubernetes/annotations_and_labels.md
mig4: Do you mean that all the sub-headings of _"Common Labels"_ should be one level up? E.g. that `### Labels Set In Custom Resources` should actually be `## Labels Set In Custom Resources`?

kopiczko: Sorry my mind was in a dark place apparently. `###` is good.

kopiczko: Updated my suggestion.

kubernetes/annotations_and_labels.md
mig4: Done

kubernetes/annotations_and_labels.md
mig4: Done

kubernetes/annotations_and_labels.md
mig4: Done

kubernetes/annotations_and_labels.md
mig4: So how about something more explicit like _Labels Set On Objects Installed By Charts_?

mig4: Went with that, let me know if that works.

kubernetes/annotations_and_labels.md
mig4: I added example of using template helper at the bottom.

kubernetes/annotations_and_labels.md
```diff
@@ -16,9 +16,10 @@ This page defines common annotations and labels we set in Kubernetes objects.
 ## Common Labels
 
 - `app` - value should contain the name of the application. Should be applied
-  to every Kubernetes resource associated with an application. This together with `version`
-  label should also be used as replicas selector. E.g. `app=kvm-operator,version=1.0.0`. Exceptions can
-  be made to accomodate for adherence to existing selectors upstream.
+  to every Kubernetes resource associated with an application. E.g.
+  `app=kvm-operator`. This is softly deprecated and should be replaced with
+  `app.kubernetes.io/name`, it's only kept to avoid breaking existing
+  workflows.
```
kopiczko: I'd put it like:

```
- `app` - should be set to the same value as `app.kubernetes.io/name`. It's only kept
  to avoid breaking existing workflows.
```

kubernetes/annotations_and_labels.md
```diff
@@ -160,6 +182,156 @@ This page defines common annotations and labels we set in Kubernetes objects.
...
+
+- `aws-operator.name` - name of the chart, trimmed to 63 characters
+- `aws-operator.chart` - normalised name + version of the chart, i.e. trimmed
+  to 63 characters and with `+` signs replaced with `-`
+- `aws-operator.labels` - defines all the labels described above, including the
+  selector labels; usage: `{{- include "aws-operator.labels" . | nindent 4 }}`
+  (adjust indent as required)
+- `aws-operator.selectorLabels` - defines labels to be used in selectors;
+  usage: `{{- include "aws-operator.selectorLabels" . | nindent 6 }}`
+  (adjust indent as required)
```
kopiczko:
```suggestion
  usage: `{{- include "aws-operator.selectorLabels" . | nindent INDENT }}`
  (set INDENT to required integer)
```

kubernetes/annotations_and_labels.md
```diff
@@ -160,6 +182,156 @@ This page defines common annotations and labels we set in Kubernetes objects.
...
+This defines a couple of strings based on chart name/version normalised to
+comply with k8s resource name [constraints][k8s-identifiers], and two snippets
+with labels:
+
+- `aws-operator.name` - name of the chart, trimmed to 63 characters
+- `aws-operator.chart` - normalised name + version of the chart, i.e. trimmed
+  to 63 characters and with `+` signs replaced with `-`
+- `aws-operator.labels` - defines all the labels described above, including the
+  selector labels; usage: `{{- include "aws-operator.labels" . | nindent 4 }}`
+  (adjust indent as required)
```
kopiczko: To avoid copy/paste errors.

```suggestion
  selector labels; usage: `{{- include "aws-operator.labels" . | nindent INDENT }}`
  (set INDENT to required integer)
```

kopiczko (commented):
This is really nice. LGTM thanks <3

kopiczko (approved):
Forgot to approve!

kubernetes/annotations_and_labels.md
mig4: Done

kubernetes/annotations_and_labels.md
mig4: Done

kubernetes/annotations_and_labels.md
mig4: Done

QuentinBisson (approved):

mig4:
We talked about this in team standup today, and I'm going to go with this in a bit and will write a post to #news-dev to make everyone aware.

kubernetes/annotations_and_labels.md
```diff
@@ -41,25 +39,46 @@ This page defines common annotations and labels we set in Kubernetes objects.
...
+  every Kubernetes resource associated with that application, i.e. all objects
+  in a chart that installs it. E.g. `app.kubernetes.io/name=kvm-operator`.
+- `app.kubernetes.io/instance` - value should be set to `{{ .Release.Name }}`
+  and is meant for differentiating between instances of the same application.
+  This together with `app.kubernetes.io/name` should be used as replicas
+  selector, e.g.
+  `app.kubernetes.io/name=kvm-operator,app.kubernetes.io/instance=kvm-operator-1.0.0`.
+- informational labels, applied to all objects in a chart that installs an app
+  or operator and so they indicate its source and instance:
+  - `app.kubernetes.io/version` - value should be the app/operator version as
```
wtty-fool: Just a thought: since https://github.com/giantswarm/gatekeeper-control-plane-rules/pull/20, `app.kubernetes.io/version` is used for gatekeeper uniqueness checks [(see here)](https://github.com/giantswarm/gatekeeper-control-plane-rules/blob/master/helm/gatekeeper-control-plane-rules/templates/uniqueoperatorversion.yaml#L16).

It would be helpful to include this information here, somewhere.

mig4: Em, that's not right is it? I mean the operators still use the `$OPERATOR.giantswarm.io/version` label for marking objects they reconcile, e.g.:

- https://github.com/giantswarm/aws-operator/blob/master/service/controller/cluster_resource_set.go#L719
- https://github.com/giantswarm/aws-operator/blob/ad94f07f4a83ff2cb31b786deddf9ce64af92197/service/controller/key/common.go#L122
- https://github.com/giantswarm/aws-operator/blob/ad94f07f4a83ff2cb31b786deddf9ce64af92197/pkg/label/label.go#L18

I missed this change in GK rules, why was that done?

kopiczko: GK rules are to make sure we don't install two deployments with the same `app.kubernetes.io/name` and `app.kubernetes.io/version`.

BTW @wtty-fool we should also rename `app` to `app.kubernetes.io/name` here https://github.com/giantswarm/gatekeeper-control-plane-rules/pull/20/files#diff-d639fe45fa938f521566f27263b98d82R17

mig4: Yeah, oh right so this is because we removed `$OPERATOR.giantswarm.io/version` from the operator's own `Deployment` object. Meaning it now compares the internal project version which is exposed as `app.kubernetes.io/version` label on its own components, to `$OPERATOR.giantswarm.io/version` set on reconciled objects. OK, makes sense, I'll update descriptions.

mig4: Done. I'm going to leave this until tomorrow so in case the wording can be improved let me know.

kopiczko (approved):
I like the changes.

kubernetes/annotations_and_labels.md
```diff
@@ -15,10 +15,8 @@ This page defines common annotations and labels we set in Kubernetes objects.
 
 ## Common Labels
 
-- `app` - value should contain the name of the application. Should be applied
-  to every Kubernetes resource associated with an application. This together with `version`
-  label should also be used as replicas selector. E.g. `app=kvm-operator,version=1.0.0`. Exceptions can
-  be made to accomodate for adherence to existing selectors upstream.
+- `app` - should be set to the same value as `app.kubernetes.io/name`. It's
+  deprecated and only kept to avoid breaking existing workflows.
```
pipo02mix: We need to announce in the releases we do in order to give customers to adapt automation

pipo02mix (approved):
this is awesome

kubasobon (approved):
Thanks

tomahawk28 (approved):
LGTM

kubernetes/annotations_and_labels.md
mig4: Sure, I'll mention that we need to do that in the ticket.

PR #103, opened 2020-03-31
Add Dependabot doc

stone-z:
Adds basic docs and example for using Dependabot for automatic dependency updates

Toward https://github.com/giantswarm/giantswarm/issues/6223 and for the good of humankind

### DIFF
--- a/github/dependabot.md
+++ b/github/dependabot.md
@@ -26,14 +26,15 @@ For instance, [here are some vulnerabilities found in the k8s.io libraries we us
 
 There are two ways to configure Dependabot:
 
-1. Through the web UI located [here](https://app.dependabot.com/accounts/giantswarm/)
-2. Using configuration files described [here](https://dependabot.com/docs/config-file/)
+1. Through [the web UI](https://app.dependabot.com/accounts/giantswarm/)
+2. Using [configuration files](https://dependabot.com/docs/config-file/)
 
 You can validate a Dependabot config file [using this validator](https://dependabot.com/docs/config-file/validator/)
 
 A sample configuration file is included below, which excludes all `k8s.io` dependencies above `0.17.0`:
 
 ```yml
+# .dependabot/config.yml
 version: 1
 update_configs:
   - package_manager: "go:modules"
@@ -43,7 +44,6 @@ update_configs:
       - match:
           dependency_name: "k8s.io/*"
           version_requirement: ">=0.17.0"
-
 ```
 
 When possible, exclude only known-breaking version changes (e.g. exclude only major versions).
--- a/github/dependabot.md
+++ b/github/dependabot.md
@@ -27,7 +27,10 @@ For instance, [here are some vulnerabilities found in the k8s.io libraries we us
 There are two ways to configure Dependabot:
 
 1. Through [the web UI](https://app.dependabot.com/accounts/giantswarm/)
-2. Using [configuration files](https://dependabot.com/docs/config-file/)
+2. Using [configuration files](https://dependabot.com/docs/config-file/) (preferred)
+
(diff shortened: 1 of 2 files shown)


github/dependabot.md
```diff
@@ -0,0 +1,50 @@
...
+```yml
+version: 1
+update_configs:
+  - package_manager: "go:modules"
+    directory: "/"
+    update_schedule: "daily"
+    ignored_updates:
+      - match:
+          dependency_name: "k8s.io/*"
+          version_requirement: ">=0.17.0"
```
rossf7: Not sure how practical the version is here. There are a large number of repos where we'll need this and then we need to update the config file in all of them.

github/dependabot.md
```diff
@@ -0,0 +1,50 @@
...
+
```
rossf7: Nit but can this whitespace go?

github/dependabot.md
```diff
@@ -0,0 +1,50 @@
...
+is an important part of securing our platform.
+
+For instance, [here are some vulnerabilities found in the k8s.io libraries we use](https://snyk.io/vuln/search?type=golang&q=k8s.io)
+
+## Configuration
+
+There are two ways to configure Dependabot:
+
+1. Through the web UI located [here](https://app.dependabot.com/accounts/giantswarm/)
+2. Using configuration files described [here](https://dependabot.com/docs/config-file/)
```
xh3b4sd: I would really like to not have `here` link names. That is pretty bad UX IMO and does not say much. I am rather a fan of pure URLs or more meaningful descriptions. I often use some part of the explanation or sentence as link title. Here `config file docs` would be a better link name for instance.

github/dependabot.md
```diff
@@ -0,0 +1,50 @@
...
+
+You can validate a Dependabot config file [using this validator](https://dependabot.com/docs/config-file/validator/)
+
+A sample configuration file is included below, which excludes all `k8s.io` dependencies above `0.17.0`:
```
xh3b4sd: Would be cool to put `.dependabot/config.yml` here somewhere near the example config.

github/dependabot.md
stone-z: Gone

github/dependabot.md
stone-z: Sure :+1:  updated

github/dependabot.md
stone-z: Yes good thinking - added

github/dependabot.md
stone-z: IMO the annoying dependabot PRs are the ones which bump a minor version which we know can't be used with the k8s version for that branch. We should still get patch updates which support our k8s version though because these include security fixes. The versions will vary across branches/repos anyway, but they change infrequently within one branch/repo

github/dependabot.md
```diff
@@ -0,0 +1,50 @@
...
+is an important part of securing our platform.
+
+For instance, [here are some vulnerabilities found in the k8s.io libraries we use](https://snyk.io/vuln/search?type=golang&q=k8s.io)
+
+## Configuration
+
+There are two ways to configure Dependabot:
+
+1. Through [the web UI](https://app.dependabot.com/accounts/giantswarm/)
+2. Using [configuration files](https://dependabot.com/docs/config-file/)
```
MarcelMue, line 30: Can we add a sentence here that the config file is the preferred way?

It would make maintaining dependabot much easier IMO.

stone-z: Good idea

github/dependabot.md
rossf7: OK cool I get the reasoning for having the version requirement here. This isn't a blocker for me. Let's try it out and see how it goes.

PR #106, opened 2020-04-08
Add info on golangci-lint

fiunchinho:
Adding help on how to use golangci-lint from intellij.


fiunchinho:
Good idea! Please take a look.

stone-z:
I would also do it in reverse -- e.g. add something about "If you use intellij, you can find tool info here" to the other page as well, so that people can find this

go/general_style_guide.md
```diff
@@ -39,3 +39,8 @@ This should be used sparingly to avoid forming a bad habit of simply silencing b
 
 Details for using all of these methods can be found in the
 [False Positives](https://github.com/golangci/golangci-lint#false-positives) section of the docs.
+
+## Tooling
```
fiunchinho, line 43: Not sure if it's the best header.

fiunchinho:
Done

PR #107, opened 2020-04-08
kubernetes: add resource names stem to `_helpers.tpl`

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/9878.

Define a common name stem for resource names in helpers template to
reduce repetition in charts' `values.yaml` file.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -225,6 +225,18 @@ Create chart name and version as used by the chart label.
 {{- end -}}
 
 {{/*
+Create a name stem for resource names
+
+When pods for deployments are created they have an additional 16 character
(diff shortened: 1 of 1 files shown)


mig4:
> I think it would be cool to include this in our [operator template](https://github.com/giantswarm/template-operator)

Definitely, it's on my todo list in giantswarm/giantswarm#9878 👍

kopiczko:
I'd like to finish discussions here https://github.com/giantswarm/opsctl/pull/596 before proceeding with this one.

TheoBrigitte:
Taking over from here:

* removed the `aws-operator` prefix from variables define in helm chart helper
* split the resource name variables into a different helper file `_resource.tpl`

PR #109, opened 2020-04-16
project_structure: add docs for Go library structure

kopiczko:


go/project_structure/library.md
```diff
@@ -0,0 +1,113 @@
...
+```
+./pkg/k8sversion
+├── error.go
+├── funcs.go
+├── funcs_test.go
+└── types.go
+```
+
+- `error.go` holds error definitions for the package.
+- `funcs.go` holds functions which are not methods of any struct (in this case
```
tuommaki, line 106: I have to say I'm not personally super thrilled to organize code into files this way. I prefer to have file per context and combine related stuff there, but I might be minority with my opinion.

go/project_structure/library.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Root package
+
+Note that since we use Go modules we don't have any Go files in the root of the
+repository. This is because as we bump the major version of the module we need
+to use `/vN/` imports. We put everything under subpackages of `/pkg` to avoid
+some ugly import aliases like:
+
+```
+k8sclientv3 "github.com/giantswarm/k8sclient/v3"
```
xh3b4sd, line 49: In order to make that work the `go.mod` file needs to define `module github.com/giantswarm/k8sclient/v3`, which would be good to document as well. See https://github.com/giantswarm/k8sclient/blob/master/go.mod#L1.

go/project_structure/library.md
kopiczko: I didn't mean enforcing here. I rather wanted to provide a default. I understand this doesn't fit everything. That's why I wrote `This is a typical layout`.

TBH it started with `/pkg` and Go modules `vN`. I'm even fine with removing whole `k8scrdclient` section if it feels "offensive".

go/project_structure/library.md
kopiczko: > This is because as we bump the major version of the module we needto use `/vN/` imports.

Isn't that the same thing? I'm not sure where to put this information with the current flow of the document. Suggestions welcome.

xh3b4sd: I forgot to get the `go.mod` file right a couple of times already. The document here does not appear to tell me that. I am not sure if it should tell me that. I just wanted to bring forth the point and it kind of fits when you already suggest to have fancy imports. Because then the question is how to make them actually work.

xh3b4sd: Also what I figured is that `module github.com/giantswarm/k8sclient/v1` does not work because given `vN` then `N >= 2`.

go/project_structure/library.md
```diff
@@ -0,0 +1,151 @@
...
+    │   ├── error.go
+    │   └── rest_config.go
+    └── k8sversion
+        ├── error.go
+        ├── funcs.go
+        ├── funcs_test.go
+        └── types.go
+```
+
+Below the go.mod file and a few packages discussed.
```
tfussell, line 39:
```suggestion
The go.mod file and packages are discussed below.
```

go/project_structure/library.md
```diff
@@ -0,0 +1,151 @@
...
+like:
+
+```
+module github.com/giantswarm/k8sclient
+```
+
+For releases `v2.x.x`:
+
+```
+module github.com/giantswarm/k8sclient/v2
```
tfussell, line 65: What happens if this is still `module github.com/giantswarm/k8sclient` and we tag the library as `v2.0.0`? It will just fail to import?

kopiczko: It will fail to import talking about incompatibility.

go/project_structure/library.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Root package
+
+Since we use Go modules we don't have any Go files in the root of the library
+repository. This is because as we bump the major version of the module we need
+to use `/vN` imports. We put everything under subpackages of `/pkg` to avoid
+some ugly import aliases like:
+
+```
+k8sclientv3 "github.com/giantswarm/k8sclient/v3"
```
xh3b4sd, line 87:
```suggestion
import k8sclientv3 "github.com/giantswarm/k8sclient/v3"
```

go/project_structure/library.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+Instead the import will look like:
+
+```
+"github.com/giantswarm/k8sclient/v3/pkg/k8sclient"
```
xh3b4sd, line 93:
```suggestion
import "github.com/giantswarm/k8sclient/v3/pkg/k8sclient"
```

go/project_structure/library.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+This comes at expense of longer imports of other packages. E.g. instead of
+having import like:
+
+```
+"github.com/giantswarm/k8sclient/v3/k8scrdclient"
```
xh3b4sd, line 100:
```suggestion
import "github.com/giantswarm/k8sclient/v3/k8scrdclient"
```

go/project_structure/library.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+We have import like:
+
+```
+"github.com/giantswarm/k8sclient/v3/pkg/k8scrdclient"
```
xh3b4sd, line 106:
```suggestion
import "github.com/giantswarm/k8sclient/v3/pkg/k8scrdclient"
```

go/project_structure/library.md
xh3b4sd: +1 on not enforcing it. The default guideline makes sense to me.

tuommaki: Sorry I was a bit ambiguous here. I meant specifically `funcs.go` (like we similarly have `error.go`, `spec.go` and `types.go`).

I perfectly understand the `/pkg` structure with Go modules `vN` approach. I'm just converting `azure-operator` to Go modules and already hit this specific import "issue" few times.

Having a guideline is good. Strict enforcing not so :slightly_smiling_face:

PR #112, opened 2020-04-23
Add microerror livetemplate for intellij

fiunchinho:
This PR adds help on how to configure IntelliJ so that it suggests
```go
if err != nil {
	return microerror.Mask(err)
}
```

when typing `err`.


go/intellij.md
```diff
@@ -35,3 +40,18 @@ Now, make sure that lint inspections are enabled
...
 It should show inspection warnings when linting fails.
+
+## Autocomplete with our microerror library
+
+You can make IntelliJ to autocomplete with [our microerror library](https://github.com/giantswarm/microerror/) when returning an error.
+Instead of manually writing this
+
+```go
+if err != nil {
+  return microerror.Mask(err)
```
kopiczko, line 51: This is me in a nutshell:

```suggestion
	return microerror.Mask(err)
```

PR #113, opened 2020-05-07
Add team label

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/10623

See discussion in the issue. We would like to add a `app.giantswarm.io/team` label for routing alerts.

The source of truth will be the `maintainers` field in Chart.yaml and `architect helm template` will need to be extended to support this.


kopiczko:
@mig4 may be interesting for you.

kubernetes/annotations_and_labels.md
```diff
@@ -205,6 +206,14 @@ and should be placed inside the chart's `templates/` subdirectory and they
 define a couple of variables that can then be re-used throughout other
 templates in the chart.
 
+The team label is set using the `maintainers` field in Chart.yaml. The prefix
+`team-` identifies it and is used to set `.Values.maintainers.team`.
+
+```
+maintainers:
+- name: team-firecracker
+```
```
kopiczko, line 215: Will `.Values.maintainers.team` work with that? `maintainers` seem to be an array.

rossf7: Yes maintainers is an array so  `architect helm template` would need to parse that.

Each app would belong to a team. This also allows having squads for optional apps or individuals as extra fields in maintainers.

There is an edge case with cluster-autoscaler which belongs to 2 teams depending on the provider. There is more discussion in the issue and we still need to find a solution for that.

kubernetes/annotations_and_labels.md
```diff
@@ -205,6 +206,14 @@ and should be placed inside the chart's `templates/` subdirectory and they
 define a couple of variables that can then be re-used throughout other
 templates in the chart.
 
+The team label is set using the `maintainers` field in Chart.yaml. The prefix
+`team-` identifies it and is used to set `.Values.maintainers.team`.
```
yasn77, line 210: I am not sure why opt to use a prefix instead of a key in a yaml map? A prefix means more parsing logic.

rossf7: We'd like to use the maintainers field in Chart.yaml but it only supports name, email and url AFAIK.

https://helm.sh/docs/topics/charts/

tomahawk28: What you mean by that is, you want to avoid presenting it like, `name: batman` but `name: team-batman`, right?

kubernetes/annotations_and_labels.md
kopiczko: Why not something like .Chart.maintainers[0].name. (You’d need to use index function iirc)

rossf7: @kopiczko Thanks I like using `.Chart.maintainers[0].name` as then we don't need to change architect and can just use Helm templating.

I'll get back to this later this week.

kubernetes/annotations_and_labels.md
yasn77: @rossf7 From the docs
> Other fields will be silently ignored.

I also checked the [schema used by Chart Testing (`ct`)](https://github.com/helm/chart-testing/blob/master/etc/chart_schema.yaml), `maintainers` is optional, but if it's set then `name` is required. We could include a `team` key in `maintainers`, this will be ignored by linting and Helm. But we would still need to fill the `name` key... So yeh, I am not 100% sure, I would still prefer using a `team` key but not sure if it really matters if we have to complete `name` anyways.

yasn77: @rossf7 From the docs

Other fields will be silently ignored.

I also checked the schema used by Chart Testing (ct), maintainers is optional, but if it's set then name is required. We could include a team key in maintainers, this will be ignored by linting and Helm. But we would still need to fill the name key... So yeh, I am not 100% sure, I would still prefer using a team key but not sure if it really matters if we have to complete name anyways.

rossf7:
Using maintainers field for this is not a good fit. There are too many exceptions like apps being owned by different teams per provider.

Batman will start work soonish on the Representation of Apps story. With this we will have separate metadata YAML files per chart and this will be used to create new appcatalogentry CRs.

This metadata YAML file will be used to define the team ownership and we are not limited to the Chart.yaml format.

PR #117, opened 2020-06-26
Explain small try catch

oponder:
Just something I came across while refactoring that I thought was worth putting into fmt.

### DIFF
--- a/javascript/small_try_catch.md
+++ b/javascript/small_try_catch.md
@@ -1,0 +1,33 @@
+# Keep your Try / Catches as small as possible
+
+This will also help you keep more happy path code to the left.
+
+The goal is to keep just the code you think might throw an exception in the
+try/catch. Don't use the "try" clause as a happy path container.
+
+Don't
+```
+let app;
+try {
+  await app = createApp();
+  new FlashMessage(`Your app is being installed`);
+  return app;
+} catch (error) {
+  showAppInstallationErrorFlashMessage();
+  throw error;
+}
+```
+
+Do:
+```
+let app;
+try {
+  await app = createApp();
+} catch (error) {
+  showAppInstallationErrorFlashMessage(name, clusterId, error);
+  throw error;
+}
+
+new FlashMessage(`Your app is being installed`);
+return app;
+```


axbarsan:
I actually am not a fan of this pattern. It's less readable, because you have to scroll around and see what mutations are done to your variable, instead of having them right in front of your eyes.

It's also a matter of performance, since we declare the variable as `undefined` first, and then we change its type, when we assign the new awaited value to it.

I made a quick benchmark illustrating that:
https://jsperf.com/declaring-vars-inside-try-block-vs-outside-of-it

image

oponder:
Ah yeah, 22% slower is a good argument.

But I do think avoiding having too many lines in a try / catch is a good idea.

What if there was anyways some more code that depended on that, and another thing that could error in a different way later. Then I think we'd have to do the `let` outside of the try/catch anyways.

Maybe we can do it by factoring out the try/catch always into a small function that all it does is return the value

Kinda want to go towards what they're discussing here:

https://medium.com/knock-engineering/happy-path-error-handling-1f698068a345

axbarsan:
> But I do think avoiding having too many lines in a try / catch is a good idea.

As a way to combat that, we can always extract functionality into other smaller functions that can throw exceptions, and construct the error handling context into a smaller function.

> What if there was anyways some more code that depended on the result (`app` in the example case), and another thing that could error in a different way later. Then I think we'd have to do the `let` outside of the try/catch anyways.

Try/catch blocks can be nested

```javascript
async function test() {
	try {
		const app = await getApp(id);

		try {
			app.config = await getAppConfig(app.configName);
		} catch (err) {
			doSomeOtherWithError(err);
		}

		return app;
	} catch(err) {
		doSomethingWithError(err);
	}
}
```

> Maybe we can do it by factoring out the try/catch always into a small function that all it does is return the value

I don't believe that's necessary. Try/catch blocks are part of the language standard and are pretty light and simple to begin with. Would we abstract variable declaration with the `let` keyword into a helper like `declareVariable()`?


I hope I'm not "cutting your wings" by saying this. I'm all for improving our code quality, but I don't think this rule is necessary.

oponder:
Yeah I think we are kind of on the same page, for me it doesn't have to be a rule, but it is kind of a specific example of the 'leftmost happy path' fmt rule that we do already have, only applied to try/catch.

I really do think having the happy path be to the left makes it much easier to parse, and to get a feeling for what a function will actually do at a quick glance. 

I would for sure want to avoid the double nested try/catch.

I think javascript has a lot of things that are part of the language that we actively avoid because they're not a great idea. Using exceptions for control flow is something Im not a big fan of personally.

I'm ok to not make this a fmt-rule though, but to still allow it in cases where it makes sense.

But anyways, I think we're kind of on the same page when you say:

>  we can always extract functionality into other smaller functions that can throw exceptions

axbarsan:
If we want to avoid nested try/catch blocks visually, we can always extract the block/functionality to a different function. In this case the new function will not throw.

This way our initial `test()` function will be small, and we would have a custom error handling logic for the `getAppConfig()` function. 

```javascript
async function test() {
	try {
		const app = await getApp(id);
		app.config = await getAppConfig(app.configName);

		return app;
	} catch(err) {
		doSomethingWithError(err);
	}
}

async function getAppConfig(configName) {
	try {
		const config = await someFetchFunction(configName);
			
		return config;
	} catch (err) {
		doSomeOtherThingWithError(err);
	}
}
```

PR #118, opened 2020-07-02
Add styled components section

ubergesundheit:
This PR adds a section to javascript rules about avoiding nested css selectors in styled components


oponder:
Merging with admin powerz. 

I'll edit my open PR to avoid nested selectors and try to avoid doing that in the future.

PR #119, opened 2020-07-06
Update dependabot docs for new version

stone-z:
Following on from [a slack thread](https://gigantic.slack.com/archives/C04TGHDEF/p1594029615188200), include sample with new syntax and upgrade instructions for dependabot.


stone-z:
> Maybe we can also mention here that we can use devctl.

Yeah good idea, I added a reference to it

github/dependabot.md
```diff
@@ -24,30 +24,55 @@ For instance, [here are some vulnerabilities found in the k8s.io libraries we us
...
 
-There are two ways to configure Dependabot:
+Dependabot updates can be enabled for a repository in the "Settings & analysis" tab of the GitHub settings UI.
+To enable updates, simply click the "Enable" button as shown in the image below.
+![](dependabot/dependabot-enable.png)
 
-1. Through [the web UI](https://app.dependabot.com/accounts/giantswarm/)
-2. Using [configuration files](https://dependabot.com/docs/config-file/) (preferred)
+To further configure how Dependabot performs updates in your repository, use a configuration file.
+You can generate a stub configuration file using `devctl gen dependabot` (see [`devctl`](https://github.com/giantswarm/devctl) for usage).
```
kopiczko, line 32: Is there a reason to maintain this file manually when we generate it with `devctl`?

@njuettner is it supposed to be a stub when the file is regenerable in your implementation?

njuettner: Ideally everything should be handled via devctl. 
Right now I only focussed on implementing it for go projects, I'm not aware how many other repos we have which use different languages.

kopiczko: Okay. @stone-z can we reflect that here? I mean that for **all** Go projects we should use what `devctl gen dependabot` generates.

stone-z: We have projects which are not Go, and we have different needs between Go projects, so IMO yes

stone-z: Sorry, previous comment was based on old GH cached state. Updated

github/dependabot.md
```diff
@@ -24,30 +24,59 @@ For instance, [here are some vulnerabilities found in the k8s.io libraries we us
...
 
-1. Through [the web UI](https://app.dependabot.com/accounts/giantswarm/)
-2. Using [configuration files](https://dependabot.com/docs/config-file/) (preferred)
+To further configure how Dependabot performs updates in your repository, use a configuration file.
 
-Configuration files are preferred because they can be more easily managed per-project and per-branch directly in the repository,
-and expose more configuration options than the web UI.
+For `Go` projects, you should generate one using `devctl gen dependabot` (see [`devctl`](https://github.com/giantswarm/devctl) for usage).
+If needed, you can then modify this file to control the update frequency or adjust dependency preferences.
+However, keep in mind that re-generating with `devctl` will overwrite any such customizations.
```
kopiczko, line 35: I would strongly suggest skipping this part. We should adapt the generating code if we find some edge cases. If this is impossible then as last resort we should do manual modifications. But let's find what are those edge cases first.

```suggestion
```

PR #123, opened 2020-07-17
Add node pool labels

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/160

What's in the box:
- `giantswarm.io/machine-pool` - value should contain tenant cluster node pool ID (i.e. the machine pool ID) which this object is part of. E.g. `giantswarm.io/machine-pool=de19f`.
- `machine-pool.giantswarm.io/name`: value contains a human-friendly node pool name set by the customer.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -39,10 +39,6 @@ This page defines common annotations and labels we set in Kubernetes objects.
   services (i.e. K8s components) or `managed` for Giant Swarm managed services
   (i.e. networking, DNS, monitoring, ingress controller, etc.). Latter would be
   managed by `chart-operator`.
-- `machine-pool.giantswarm.io/subnet`: value should contain the name of the
-  subnet where the tenant cluster node pool is deployed. This name should be
-  equivalent to the tenant cluster node pool ID. E.g.
-  `machine-pool.giantswarm.io/subnet=de19f-1`.
 - `release.giantswarm.io/version` - value should be Giant Swarm release
   version, e.g. `release.giantswarm.io/version=2.3.0`.
 - `OPERATOR.giantswarm.io/version` - value should be the version of the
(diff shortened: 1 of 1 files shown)


nprokopic:
@tuommaki @fiunchinho did we decide what will be the final format for the node pool name and its subnet name?

xh3b4sd:
We recently introduced https://github.com/giantswarm/apiextensions/tree/master/pkg/label and thought it would be good to have a programmatic source of truth. We want to start using the labels in `apiextensions` in the operators and for docs generation. So it would be cool if we could start adding labels in the go package. We can link from here to there so people find what they are looking for. I actually forgot to post a news message about it. Is everyone cool with that new approach? Is there more to talk about?

whites11 (approved):

kubernetes/annotations_and_labels.md
```diff
@@ -36,6 +39,10 @@ This page defines common annotations and labels we set in Kubernetes objects.
   services (i.e. K8s components) or `managed` for Giant Swarm managed services
   (i.e. networking, DNS, monitoring, ingress controller, etc.). Latter would be
   managed by `chart-operator`.
+- `machine-pool.giantswarm.io/subnet`: value should contain the name of the
```
fiunchinho, line 42: I think this is not needed. The subnet can be called like the node pool it was created for. Who can we talk to so we can discuss if it could be changed in AWS?

nprokopic:
> We want to start using the labels in apiextensions in the operators and for docs generation. So it would be cool if we could start adding labels in the go package. We can link from here to there so people find what they are looking for.

:+1:  I'll then add new machine pool labels to `apiextensions` and link them here, and then use them in the `API` as well (where we are using them currently).

> Is everyone cool with that new approach? Is there more to talk about?

All good, thanks for getting us up-to-date with this.

kubernetes/annotations_and_labels.md
nprokopic: My thinking was that it is easier that we add it for Azure now, as we are in development, than to change it for AWS which is already working this way for some time.

But I guess we could skip using the label for subnet anyway, if we are to follow.

P.S. We could maybe check out if we can have node pool subnet naming compatible with CAPZ.

nprokopic: Subnet annotation is removed now.

nprokopic:
Links to `apiextensions` where the labels and annotations are defined are added.

fiunchinho (approved):

rossf7 (approved):
LGTM

PR #125, opened 2020-08-14
Add labels and annotations used for Cluster and AzureCluster CRs

nprokopic:
These are labels and annotations that are currently planned to be used for `Cluster` and `AzureCluster` CRs.

### Note and question about `Cluster` CR labels

`Cluster` CR is used for both Azure and AWS tenant clusters and, if I'm not mistaken, for AWS clusters the `cluster.x-k8s.io/cluster-name` label is not set. Should we keep it that way and clarify the difference in the `fmt` repo, or should we make Azure and AWS consistent here?

Motivation for using upstream Cluster API labels: In azure-operator we started using some utility functions from the upstream CAPI and CAPZ projects, and those are often looking for and using upstream labels. We decided that it would be better to use label definitions from the upstream packages, than to copy them to our projects.

### DIFF
--- a/kubernetes/annotations_and_labels.md
+++ b/kubernetes/annotations_and_labels.md
@@ -4,6 +4,8 @@ This page defines common annotations and labels we set in Kubernetes objects.
 
 ## Common Annotations
 
+- `cluster.giantswarm.io/description` - value contains a human-friendly cluster
+  name set by the customer. It is defined in [`github.com/giantswarm/apiextensions/pkg/annotation` package](https://github.com/giantswarm/apiextensions/blob/master/pkg/annotation/cluster.go).
 - `giantswarm.io/docs` - value should be an URL to a common documentation
   location. For now this should be in our `giantswarm/giantswarm` repository in
   which we crosslink all pages to create a reasonable documentation of all kinds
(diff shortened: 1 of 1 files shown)


fiunchinho:
AFAIC the plan is to eventually end up using CAPI in AWS. So I'd make it consistent across AWS and Azure.

kubernetes/annotations_and_labels.md
```diff
@@ -133,6 +135,26 @@ querying by shared tooling.
 
 ##### Labels
 
+- `cluster.x-k8s.io/cluster-name`
```
fiunchinho, line 138: IIUC it would contain whatever is in `Meta.Name`, and it's just added as label as it's more convenient when filtering. We can maybe add that as a short description.

nprokopic: Added label description at the top of this page, alongside other label descriptions.

fiunchinho (approved):

tuommaki (approved):
:+1:

kubernetes/annotations_and_labels.md
```diff
@@ -19,6 +21,10 @@ This page defines common annotations and labels we set in Kubernetes objects.
 
 - `app` - should be set to the same value as `app.kubernetes.io/name`. It's
   deprecated and only kept to avoid breaking existing workflows.
+- `cluster.x-k8s.io/cluster-name` - currently set to same value as
```
xh3b4sd, line 24: Is this only the equivalent to `giantswarm.io/cluster`? Will we put `cluster.x-k8s.io/cluster-name` also in all CRs associated with the cluster like we do with `giantswarm.io/cluster`?

tuommaki: Yes, that would be the idea. We should also start widely using `OwnerReferences` correspondingly, though that's slightly different thing.

xh3b4sd (approved):

kubernetes/annotations_and_labels.md
nprokopic: Specifically we will put `cluster.x-k8s.io/cluster-name` to `MachinePool` and `AzureMachinePool` as well (and I believe `AzureMachine` used for the master nodes). I'll open another `fmt` PR for this (just to clarify first few other details around `MachinePool`/`AzureMachinePool` labels and annotations).

nprokopic:
Thanks folks for your feedback! :heart:

PR #126, opened 2020-08-18
Add section about nancy to Go guide

stone-z:
Nancy is a vulnerability scanner for Go dependencies. It will be added to `architect-orb` by https://github.com/giantswarm/architect-orb/pull/173

This section includes some info about what to expect and how to work with it

### DIFF
--- a/go/general_style_guide.md
+++ b/go/general_style_guide.md
@@ -78,9 +78,25 @@ A finding from `nancy` might look like:
 
 ```
 
-The best course of action is to update the dependency.
-If the dependency is nested, try using a `replace` directive in `go.mod` to use a safe version.
-If it is not possible to update the dependency for technical or compatibility reasons, you can suppress the finding by adding it to a file named `.nancy-ignore` in the root of your repository.
+For each identified vulnerability:
+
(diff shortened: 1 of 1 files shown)


stone-z:
Pinging @giantswarm/chapter-pe as this will eventually apply to all Go code

xh3b4sd:
The advice is to ignore vulnerabilities. IMO when we do that we don't need the integration at all. At which point should we escalate and maybe look for a better alternative of the vulnerable dependency? When would we say we try to fix it ourselves? How do we want to measure impact?

stone-z:
> The advice is to ignore vulnerabilities.

Of course not, the advice is to 1. update the dependency, 2.override it using `replace` if you can't update it directly, and then 3. ignore if it cannot be updated. I will add some text to highlight that.

xh3b4sd:
I understand. Though, looking down the road, the consequence will be that we start to ignore reports. How to deal with that?

corest (approved):

stone-z:
I'd like to proceed with this, @xh3b4sd please let me know if you think there should be something else included here

PR #130, opened 2020-09-22
Add Helm checksum annotations

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/13309

Tnis proposes changing the Helm release revision to checksum annotations as recommended in the Helm docs.
https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

This will reduce the number of rolled pods when updating draughtsman values or if there is a problem with reconciling app CRs.

### DIFF
--- a/helm/checksum_annotations.md
+++ b/helm/checksum_annotations.md
@@ -12,7 +12,7 @@ annotations are not needed.
 
 ## Annotations
 
-- The secret annotation is only needed if chart has a templated secret.
+- The secret annotation is only needed if the chart has a templated secret.
 
 ```yaml
 apiVersion: apps/v1
--- a/helm/checksum_annotations.md
+++ b/helm/checksum_annotations.md
@@ -12,7 +12,7 @@ annotations are not needed.
 
 ## Annotations
 
-- The secret annotation is only needed if Chart has a template.
+- The secret annotation is only needed if chart has a templated secret.
 
 ```yaml
 apiVersion: apps/v1
--- a/helm/checksum_annotations.md
+++ b/helm/checksum_annotations.md
@@ -1,0 +1,43 @@
+# Helm Checksum Annotations
+
+## Overview
+
+- We add checksum annotations to our deployments and daemonsets to trigger
+fresh pods when their configmaps or secrets have changed.
+- This reduces how often we need to roll pods which can be disruptive.
+- Changes to other resources like PSP and RBAC are applied without needing to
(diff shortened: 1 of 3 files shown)


helm/checksum_annotations.md
```diff
@@ -0,0 +1,43 @@
...
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    metadata:
+      annotations:
+        app.giantswarm.io/config-checksum: {{ include (print .Template.BasePath "/configmap.yaml") . | sha256sum | quote }}
+        app.giantswarm.io/secret-checksum: {{ include (print .Template.BasePath "/secret.yaml") . | sha256sum | quote }}
```
rossf7, line 25: This is basically the same as the Helm docs I just added the prefix.

helm/checksum_annotations.md
```diff
@@ -0,0 +1,43 @@
+# Helm Checksum Annotations
+
+## Overview
+
+- We add checksum annotations to our deployments and daemonsets to trigger
+fresh pods when their configmaps or secrets have changed.
+- This reduces how often we need to roll pods which can be disruptive.
+- Changes to other resources like PSP and RBAC are applied without needing to
```
rossf7, line 8: From the discussion earlier https://gigantic.slack.com/archives/C3C7ZQXC1/p1600771250010300
we think we only need annotations for CMs and secrets.

If there is a case we're missing please chime in.

helm/checksum_annotations.md
```diff
@@ -0,0 +1,43 @@
...
+roll pods.
+- If the app has auto reload functionality like Nginx Ingress Controller these
+annotations are not needed.
+
+## Annotations
+
+- The secret annotation is only needed if Chart has a template.
```
MarcelMue, line 15: I don't understand this sentence - can you rephrase it? Does it mean if the chart has a templated secret?

rossf7: Yes that's what I meant :) but this is much better.

MarcelMue (approved):
Works for me - interested what others think :D

tomahawk28 (approved):
LGTM

puja108 (approved):

rossf7:
Any more opinions? :) Otherwise I'll merge this afternoon and do a #news-dev post.

mig4 (approved):
Nice!

PR #131, opened 2020-11-17
Add How To for global gitleaks pre-commit check

stone-z:
Adds a short how to for globally enabling client-side gitleaks pre-commit hooks.

### DIFF
--- a/go/general_style_guide.md
+++ b/go/general_style_guide.md
@@ -127,7 +127,19 @@ You can run `nancy` locally by building or downloading the binary and running `g
 
 Please don't push secrets to GitHub unless they are encrypted.
 
-We recommend [configuring a client-side git hook](https://github.com/zricethezav/gitleaks/wiki/Scanning#uncommitted-changes-scan) so that your secrets are never pushed in the first place.
+We recommend [configuring a client-side git hook](https://github.com/zricethezav/gitleaks/wiki/Scanning#uncommitted-changes-scan) so that your secrets are never pushed in the first place. To do this:
+```
+# Create a directory to hold your global git hooks, if you don't have one
+$ mkdir /path/to/your/git-hooks
+
+# Create a file named pre-commit in that directory, with the content from the gitleaks link above
+
+# Add the path as your global git hook location
+$ git config --global core.hooksPath /path/to/your/git-hooks
+
+# Enable gitleaks globally
+$ git config --global hooks.gitleaks true
+```
 
 Just in case someone accidentally pushes a secret, there is a [`gitleaks`](https://github.com/zricethezav/gitleaks/) workflow included in `devctl gen workflows` which will scan your commits and fail your build if a suspected secret is found.
 
--- a/go/general_style_guide.md
+++ b/go/general_style_guide.md
@@ -132,7 +132,31 @@ We recommend [configuring a client-side git hook](https://github.com/zricethezav
 # Create a directory to hold your global git hooks, if you don't have one
 $ mkdir /path/to/your/git-hooks
 
-# Create a file named pre-commit in that directory, with the content from the gitleaks link above
+# Change to the directory
+$ cd /path/to/your/git-hooks
+
+# Create a file named pre-commit
+$ touch pre-commit
+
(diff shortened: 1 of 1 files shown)


go/general_style_guide.md
```diff
@@ -127,7 +127,19 @@ You can run `nancy` locally by building or downloading the binary and running `g
 
 Please don't push secrets to GitHub unless they are encrypted.
 
-We recommend [configuring a client-side git hook](https://github.com/zricethezav/gitleaks/wiki/Scanning#uncommitted-changes-scan) so that your secrets are never pushed in the first place.
+We recommend [configuring a client-side git hook](https://github.com/zricethezav/gitleaks/wiki/Scanning#uncommitted-changes-scan) so that your secrets are never pushed in the first place. To do this:
+```
+# Create a directory to hold your global git hooks, if you don't have one
+$ mkdir /path/to/your/git-hooks
+
+# Create a file named pre-commit in that directory, with the content from the gitleaks link above
```
MarcelMue, line 135: Can we add that in here directly? shouldn't be too hard to `touch` a file and then pipe that stuff in

stone-z: Yeah I can include it here, my thinking was that this might get out of date with upstream then

MarcelMue (approved):

go/general_style_guide.md
```diff
@@ -127,7 +127,43 @@ You can run `nancy` locally by building or downloading the binary and running `g
 
 Please don't push secrets to GitHub unless they are encrypted.
 
-We recommend [configuring a client-side git hook](https://github.com/zricethezav/gitleaks/wiki/Scanning#uncommitted-changes-scan) so that your secrets are never pushed in the first place.
+We recommend [configuring a client-side git hook](https://github.com/zricethezav/gitleaks/wiki/Scanning#uncommitted-changes-scan) so that your secrets are never pushed in the first place. To do this:
+```
```
kopiczko, line 131:
```suggestion
```bash
```

kopiczko (approved):

go/general_style_guide.md
```diff
@@ -127,7 +127,43 @@ You can run `nancy` locally by building or downloading the binary and running `g
...
+$ touch pre-commit
+
+# Add the git hook content from the gitleaks wiki
+$ cat <<EOS >> pre-commit
+#!/bin/sh
+# Adds gitleaks as a pre-commit hook.
+
+gitleaksEnabled=\$(git config --bool hooks.gitleaks)
+cmd="gitleaks --verbose --redact --pretty"
+if [ "\$gitleaksEnabled" = "true" ]; then
```
kopiczko, line 148: I don't think it will work if you don't run `git config hooks.gitleaks true` first

```suggestion
if [ "\$gitleaksEnabled" != "false" ]; then
```

go/general_style_guide.md
```diff
@@ -127,7 +127,43 @@ You can run `nancy` locally by building or downloading the binary and running `g
...
+    \$cmd
+    if [ \$? -eq 1 ]; then
+cat <<\EOF
+Error: gitleaks has detected sensitive information in your changes.
+If you know what you are doing you can disable this check using:
+    git config hooks.gitleaks false
+EOF
+exit 1
```
kopiczko, line 156: To make it look (hopefully) less awful:

```suggestion
        echo >&2 "Error: gitleaks has detected sensitive information in your changes."
        echo >&2 "If you know what you are doing you can disable this check using:"
        echo >&2 "      git config hooks.gitleaks false"
        exit 1
```

go/general_style_guide.md
stone-z: If the user hasn't set it then the value is just an empty string, not false

go/general_style_guide.md
stone-z: I'd like to keep the script the same as the upstream one just to reduce the overhead of keeping it updated

kopiczko: I didn't know it's upstream script. Proceed as you see the fit.

ericgraf (approved):

